PR #52, opened 2015-06-27
Chuongv/audio tests and demo

Chuongv:
Demo has been implemented. Feel free to test it out :smile: 

Thanks

### DIFF
--- a/Classes/Private/Manager/Database/OCTRealmManager.m
+++ b/Classes/Private/Manager/Database/OCTRealmManager.m
@@ -52,7 +52,7 @@
         _realm = [RLMRealm realmWithPath:path];
     });
 
-    [self removeAllCalls];
+    [self convertAllCallsToMessages];
 
     return self;
 }
@@ -260,7 +260,7 @@
     });
 }
 
-- (void)removeAllCalls
+- (void)convertAllCallsToMessages
 {
     RLMResults *calls = [OCTCall objectsInRealm:self.realm where:nil];
 
@@ -269,8 +269,7 @@
     RBQRealmChangeLogger *logger = [self logger];
 
     for (OCTCall *call in calls) {
-        OCTMessageCallEvent event = (call.status != OCTCallStatusActive) ? OCTMessageCallEventUnanswered : OCTMessageCallEventAnswered;
-        [self addMessageCallEvent:event call:call callDuration:call.callDuration];
+        [self addMessageCall:call];
     }
 
     [self.realm beginWriteTransaction];
@@ -311,15 +310,23 @@
     return messageAbstract;
 }
 
-- (void)addMessageCallEvent:(OCTMessageCallEvent)event
-                       call:(OCTCall *)call
-               callDuration:(NSTimeInterval)duration
+- (void)addMessageCall:(OCTCall *)call
 {
     NSParameterAssert(call);
     DDLogInfo(@"OCTRealmManager: adding messageCall to call %@", call);
 
+    OCTMessageCallEvent event;
+    switch (call.status) {
+        case OCTCallStatusDialing:
+        case OCTCallStatusRinging:
+            event = OCTMessageCallEventUnanswered;
+            break;
+        case OCTCallStatusActive:
+            event = OCTMessageCallEventAnswered;
+    }
+
     OCTMessageCall *messageCall = [OCTMessageCall new];
-    messageCall.callDuration = duration;
+    messageCall.callDuration = call.callDuration;
     messageCall.callEvent = event;
 
     OCTMessageAbstract *messageAbstract = [OCTMessageAbstract new];
--- a/Classes/Private/Manager/Database/OCTRealmManager.m
+++ b/Classes/Private/Manager/Database/OCTRealmManager.m
@@ -323,6 +323,7 @@
             break;
         case OCTCallStatusActive:
             event = OCTMessageCallEventAnswered;
+            break;
     }
 
     OCTMessageCall *messageCall = [OCTMessageCall new];
--- a/Classes/Private/Manager/Database/OCTRealmManager.m
+++ b/Classes/Private/Manager/Database/OCTRealmManager.m
@@ -52,6 +52,8 @@
         _realm = [RLMRealm realmWithPath:path];
     });
 
+    [self removeAllCalls];
+
     return self;
 }
 
@@ -258,6 +260,25 @@
     });
 }
 
+- (void)removeAllCalls
+{
+    RLMResults *calls = [OCTCall objectsInRealm:self.realm where:nil];
+
+    DDLogInfo(@"OCTRealmManager: removing %lu calls", calls.count);
+
+    RBQRealmChangeLogger *logger = [self logger];
+
+    for (OCTCall *call in calls) {
+        OCTMessageCallEvent event = (call.status != OCTCallStatusActive) ? OCTMessageCallEventUnanswered : OCTMessageCallEventAnswered;
+        [self addMessageCallEvent:event call:call callDuration:call.callDuration];
+    }
+
+    [self.realm beginWriteTransaction];
+    [logger willDeleteObjects:calls];
+    [self.realm deleteObjects:calls];
+    [self.realm commitWriteTransaction];
+}
+
 - (OCTMessageAbstract *)addMessageWithText:(NSString *)text
                                       type:(OCTToxMessageType)type
                                       chat:(OCTChat *)chat
@@ -290,9 +311,9 @@
     return messageAbstract;
 }
 
-- (void)addMessageCall:(OCTMessageCallEvent)event
-                  call:(OCTCall *)call
-          callDuration:(NSTimeInterval)duration
+- (void)addMessageCallEvent:(OCTMessageCallEvent)event
+                       call:(OCTCall *)call
+               callDuration:(NSTimeInterval)duration
 {
     NSParameterAssert(call);
     DDLogInfo(@"OCTRealmManager: adding messageCall to call %@", call);
--- a/Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
+++ b/Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
@@ -8,7 +8,7 @@
 
 #import "OCTSubmanagerCalls+Private.h"
 
-const OCTToxAVAudioBitRate kDefaultAudioBitRate = 48;
+const OCTToxAVAudioBitRate kDefaultAudioBitRate = OCTToxAVAudioBitRate48;
 const OCTToxAVAudioBitRate kDefaultVideoBitRate = 400;
 
 @interface OCTSubmanagerCalls () <OCTToxAVDelegate>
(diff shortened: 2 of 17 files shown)


dvor:
This test is failing sometimes. Maybe there is something race condition and toxAV is getting freed too early. Could you check that please? If nothing comes to your head you can simply add logging in test, so next time we would see what's happening there. 

```
Failures:
  0) -[OCTToxAVTests testVersionMethods] (objcToxTests.xctest)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
/Users/travis/build/Antidote-for-Tox/objcTox/objcToxTests/OCTToxAVTests.m:478: ((toxAV.toxAV == cToxAV) is true) failed:
475 {
476     OCTToxAV *toxAV = [(__bridge OCTToxAVTests *)refToSelf toxAV];
477 
478     CCCAssertTrue(toxAV.toxAV == cToxAV);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
479 }
480 
```

Chuongv:
Looking into it.. I believe it's from the testStartAndStopMethod. When you `dispatch_source_cancel`, it does not interrupt any leftover `tox_iterate` in the queue.

> Cancellation prevents any further invocation of the event handler block for the specified dispatch source, but does not interrupt an event handler block that is already in progress

`toxav` goes to nil in the `dealloc` and then tox_iterate runs. Not sure what the best way to work around this is? We can add a cancel completion block to the `dispatch_source`?

Chuongv:
Okay, I think I solved it now. Was much simpler than I originally thought. Or at least I hope. :smile:

dvor:
`pod install` failed for some reason, I've restarted travis.

dvor:
Hm, it seems that mannol has deleted his user from github :open_mouth:

Chuongv:
Hmm that's not good.... hopefully there's some explanation for this. Right now no one on IRC knows..

Chuongv:
Okay finally got it to pass again :sweat_smile: Mannol added the old group audio chats but it was causing problems with the build, so I had to revert it back for now.

Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -258,6 +260,25 @@ - (void)removeChatWithAllMessages:(OCTChat *)chat
     });
 }
 
+- (void)removeAllCalls
```
dvor: It would be better to rename method to something more explicit like `convertAllCallsToMessages`

Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -258,6 +260,25 @@ - (void)removeChatWithAllMessages:(OCTChat *)chat
...
+{
+    RLMResults *calls = [OCTCall objectsInRealm:self.realm where:nil];
+
+    DDLogInfo(@"OCTRealmManager: removing %lu calls", calls.count);
+
+    RBQRealmChangeLogger *logger = [self logger];
+
+    for (OCTCall *call in calls) {
+        OCTMessageCallEvent event = (call.status != OCTCallStatusActive) ? OCTMessageCallEventUnanswered : OCTMessageCallEventAnswered;
```
dvor: It would look nicer if it would be positive statement

```
OCTMessageCallEvent event = (call.status == OCTCallStatusActive) ? OCTMessageCallEventAnswered : OCTMessageCallEventUnanswered; 
```

dvor: Even better it would be to use switch here. This way in case if new value will be added to enum compiler will give you an error.

```
switch (call.status) {
    case OCTCallStatusRinging:
    case OCTCallStatusDialing:
        event = OCTMessageCallEventUnanswered;
        break;
    case OCTCallStatusActive:
        event = OCTMessageCallEventAnswered;
        break;
}
```

dvor: Also as `event` and `callDuration` depends on call, there is no need to path them into `addMessageCall...` method. You can move all logic inside and simply create method like

```
addMessageFromCall:(OCTCall *)call;
```

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,12 +296,24 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
 
 - (void)toxAV:(OCTToxAV *)toxAV audioBitRateChanged:(OCTToxAVAudioBitRate)bitrate stable:(BOOL)stable friendNumber:(OCTToxFriendNumber)friendNumber
 {
-    // Lower bitrate if unstable?
+    NSArray *validBitrates = @[@8, @16, @24, @32, @48];
```
dvor: Why are you sure that these are valid bitrates and that toxav will return them?

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,12 +296,24 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
...
+
+    if (! stable) {
```
dvor: Early return would look nicer (at the start of method, before validBitrates array)

```
if (stable) {
    return;
}
```

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,12 +296,24 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
...
+        NSUInteger currentIndexBitRate = [validBitrates indexOfObject:[NSNumber numberWithInt:bitrate]];
+
+        if (currentIndexBitRate == NSNotFound) {
+            return;
+        }
+
+        NSNumber *newBitrate = [validBitrates objectAtIndex:currentIndexBitRate - 1];
```
dvor: Crash on `currentIndexBitRate == 0`

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,12 +296,24 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
...
+
+        [self.toxAV setAudioBitRate:newBitrate.intValue force:NO forFriend:friendNumber error:nil];
+    }
```
dvor, line 317: So, if bitrate is unstable we will lower it. Any ideas of possibilities of increasing bitrate, e.g. if network quality increased?

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,12 +296,24 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
...
 }
 
 - (void)toxAV:(OCTToxAV *)toxAV videoBitRateChanged:(OCTToxAVVideoBitRate)bitrate friendNumber:(OCTToxFriendNumber)friendNumber stable:(BOOL)stable
 {
-    // Lower bitrate if unstable?
+    if (! stable) {}
```
dvor: Early return. Or remove this at all until video implementation.

Classes/Private/Wrapper/OCTToxAV.m
```diff
@@ -494,6 +494,9 @@ - (void)fillError:(NSError **)error withCErrorSendFrame:(TOXAV_ERR_SEND_FRAME)cE
             code = OCTToxAVErrorSendFrameInvalid;
             failureReason = @"One of the frame parameters was invalid. E.g. the resolution may be too small or too large, or the audio sampling rate may be unsupported";
             break;
+        case TOXAV_ERR_SEND_FRAME_PAYLOAD_TYPE_DISABLED:
+            code = OCTToxAVErrorSendFramePayloadTypeDisabled;
+            failureReason = @"Either friend turned off audio or video receiving or we turned off sending for the said payload.";
```
dvor: `audio or video` -> `audio/video` looks more readable to me.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 317: Yea I'm not sure what the best way to do that is? We can submit a new non-forceful bitrate change request every few minutes in the call

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv: They were working when I was testing them, well at least for 24/32/48. I never got down lower than 24. These are just numbers that mannol suggested me to try.

dvor: Hm, can you maybe make them constants? Enum with these values will be fine.

Classes/Private/Manager/Database/OCTRealmManager.m
Chuongv: Yea this cleans up the code quite a bit. I like :+1:  Thanks

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
dvor, line 317: I think we can leave it for now. Currently it is more important just to make things work. I've created an issue so we won't forget about that https://github.com/Antidote-for-Tox/objcTox/issues/53

Chuongv:
Updated, I'm not satisfied with how I did the enums, so feel free to give me any better suggestions

Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -290,15 +310,23 @@ - (OCTMessageAbstract *)addMessageWithText:(NSString *)text
...
     DDLogInfo(@"OCTRealmManager: adding messageCall to call %@", call);
 
+    OCTMessageCallEvent event;
+    switch (call.status) {
+        case OCTCallStatusDialing:
+        case OCTCallStatusRinging:
+            event = OCTMessageCallEventUnanswered;
+            break;
+        case OCTCallStatusActive:
+            event = OCTMessageCallEventAnswered;
```
dvor, line 325: Please add `break;` here. This may save us from error in future.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -298,13 +290,37 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
...
+        case OCTToxAVAudioBitRate32:
+            newBitrate = OCTToxAVAudioBitRate24;
+            break;
+        case OCTToxAVAudioBitRate24:
+            newBitrate = OCTToxAVAudioBitRate16;
+            break;
+        case OCTToxAVAudioBitRate16:
+            newBitrate = OCTToxAVAudioBitRate8;
+            break;
+    }
```
dvor, line 317: This looks okay, you can just change the ordering to be more consistent. From lowest to highest.

Chuongv:
Thanks for the comments, just updated

dvor:
With your branches in Antidote-for-Tox project I can see all updates in my feed. So no need to comment after updating now. :-)

PR #56, opened 2015-07-04
Some extra changes for objcTox

Chuongv:
Some small changes for `OCTCall` and `OCTMessageCall`.

### DIFF
--- a/Classes/Private/Manager/Database/OCTRealmManager.m
+++ b/Classes/Private/Manager/Database/OCTRealmManager.m
@@ -332,12 +332,12 @@
     OCTMessageCall *messageCall = [OCTMessageCall new];
     messageCall.callDuration = call.callDuration;
     messageCall.callEvent = event;
-    messageCall.isOutgoing = call.isOutgoing;
 
     OCTMessageAbstract *messageAbstract = [OCTMessageAbstract new];
     messageAbstract.dateInterval = [[NSDate date] timeIntervalSince1970];
     messageAbstract.chat = call.chat;
     messageAbstract.messageCall = messageCall;
+    messageAbstract.sender = call.caller;
 
     [self addObject:messageAbstract];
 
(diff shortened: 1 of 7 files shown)


Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -332,6 +332,7 @@ - (void)addMessageCall:(OCTCall *)call
     OCTMessageCall *messageCall = [OCTMessageCall new];
     messageCall.callDuration = call.callDuration;
     messageCall.callEvent = event;
+    messageCall.isOutgoing = call.isOutgoing;
```
dvor: Instead of adding `isOutgoing` to OCTMessageCall you should set `sender` property of OCTMessageAbstract (to friend or to nil, if message is outgoing).

I suggest you
- remove `isOutgoing` from OCTCall
- add `caller` property instead (which will be nil in case of outgoing call)
- add `isOutgoing` method as it is done is OCTMessageAbstract
- when creating message from call simply assign `caller` to `sender`

This way we would have similar interface in messages and calls. Also for the future we would already have `caller` when we'll be implementing group calls.

Classes/Public/Manager/Objects/OCTMessageCall.h
```diff
@@ -21,4 +21,9 @@
  **/
 @property  OCTMessageCallEvent callEvent;
 
+/**
+ * Indicates if the call was outgoing or incoming.
+ **/
+@property BOOL isOutgoing;
```
dvor: No need for this property anymore

Classes/Public/Manager/Objects/OCTCall.h
```diff
@@ -28,6 +28,12 @@
 @property OCTCallStatus status;
 
 /**
+ * The friend who started the call.
+ * Nil if the you started the call yourself.
+ **/
+@property OCTFriend *caller;
+
+/**
```
dvor, line 36: You can also add `isOutgoing` method to OCTCall, which would just check for a caller

```
- (BOOL)isOutgoing
{
    return (self.caller == nil);
}
```

dvor:
Cool

PR #61, opened 2015-07-17
Add onHoldStartInterval

Chuongv:
Let me know if I missed anything. Thanks :+1:


Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -247,6 +247,10 @@ - (void)updateCall:(OCTCall *)call withStatus:(OCTCallStatus)status
 
     [realmManager updateObject:call withBlock:^(OCTCall *callToUpdate) {
         callToUpdate.status = status;
+
+        if (status == OCTCallStatusPaused) {
+            callToUpdate.onHoldStartInterval = [[NSDate date] timeIntervalSince1970];
+        }
```
dvor: When status is changed to not paused, you can change startInterval back to 0

PR #66, opened 2015-07-21
Reduce buffer size and cut tox_iterate by 1/20

Chuongv:
This makes the sound crystal clear. Hopefully won't be here long term.

### DIFF
--- a/Classes/Private/Manager/Audio/OCTAudioEngine.m
+++ b/Classes/Private/Manager/Audio/OCTAudioEngine.m
@@ -12,7 +12,7 @@
 
 static const AudioUnitElement kInputBus = 1;
 static const AudioUnitElement kOutputBus = 0;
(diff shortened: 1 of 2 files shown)


dvor:
Sound from Antidote to Antidote? What about other clients?

Chuongv:
Sounds great on old Antidote (1/4) to objcToxDemo (1/20). Going to test it with other clients soon hopefully

dvor:
If you like I can create wiki page with information about sound (and video in the future) quality with other clients (like the gist you've given me before). You will be able to update it with most recent information.

Chuongv:
Oh good idea!

dvor:
[Here it is](https://github.com/Antidote-for-Tox/objcTox/wiki/Audio-status). I've copied old status from your gist, you can update it with `crystal clear` sound :-)

PR #83, opened 2015-07-31
Start of video

Chuongv:
Not much of implementation has been done. However I wanted to review the overall architecture before I go any further with the implementation.

Here is my plan below:

`OCTCall` has `videoIsEnabled` property. This lets `OCTSubmanagerCalls` know whether or not if the videoEngine needs to be started with any call. We can not go off the `friendAcceptingVideo` property of `OCTCall` since the user still might not want to send video.

Video preview is only provided when the user is sending video.

`OCTVideoViewController` will hold onto the `OCTVideoView`, this class will be responsible for processing video data from `OCTVideoEngine`. So video data will flow from ToxAV-> OCTToxAV -> OCTSubmanagerCalls -> OCTVideoEngine -> OCTVideoViewController. 

Processing Incoming Video:

Starting or stopping a video engine will have nothing to do with processing incoming video. They are two separate processes.

Once we receive YUV data and `processIncomingVideo` is `YES`, we can create a `CVPixelBuffer` which then gets created as a `CIImage` that is then passed to `OCTVideoViewController` for display. From researching online, that seems to be the best way for displaying of video. Just need to find some examples of this.


Classes/Public/Manager/Submanagers/OCTSubmanagerCalls.h
```diff
@@ -45,6 +45,16 @@
...
 
 /**
+ * Enable video calling for an active call.
+ * Use this when you started a call without video in the first place.
+ * @param enable YES to enable video, NO to stop video sending.
+ * @param call Call to enable video for.
+ * @param error Pointer to an error object.
+ * @return YES on success, otherwise NO.
+ */
+- (BOOL)enable:(BOOL)enable videoSendingForCall:(OCTCall *)call error:(NSError **)error;
```
dvor: It would be better to rename this method. It is a little bit confusing now. I think that "first argument part" `enable:` should describe what the method is doing. While all other parts just describe other arguments.

```
- (BOOL)enableVideoSending:(BOOL)enable forCall:(OCTCall *)call error:(NSError **)error;
```

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +93,43 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
     return nil;
 }
 
+- (BOOL)enable:(BOOL)enable videoSendingForCall:(OCTCall *)call error:(NSError **)error
+{
+    BOOL success = [self setVideoBitrate:kDefaultVideoBitRate forCall:call error:error];
```
dvor: Why would you set bitrate when disabling video?

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +93,43 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
...
+
+    if (! success) {
+        return NO;
+    }
+
+    if (enable && (call.pausedStatus == OCTCallPausedStatusNone)) {
```
dvor: `(call.pausedStatus == OCTCallPausedStatusNone)` is used in several places. Maybe we can create category (for objcTox only, not for end user) for OCTCall with `- (BOOL)isPaused` method?

We can make this method public but I think that OCTCall API will be getting too complicated this way.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +93,43 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
...
+        self.videoEngine.friendNumber = friend.friendNumber;
+        [self.videoEngine startSendingVideo];
+    }
+    else {
+        [self.videoEngine stopSendingVideo];
+    }
+
+    OCTRealmManager *manager = [self.dataSource managerGetRealmManager];
+    [manager updateObject:call withBlock:^(OCTCall *callToUpdate) {
+        call.videoIsEnabled = enable;
```
dvor: `call` -> `callToUpdate`

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +93,43 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
...
+    }];
+
+    return success;
+}
+
 - (BOOL)answerCall:(OCTCall *)call enableAudio:(BOOL)enableAudio enableVideo:(BOOL)enableVideo error:(NSError **)error
 {
     OCTToxAVAudioBitRate audioBitRate = (enableAudio) ? kDefaultAudioBitRate : OCTToxAVAudioBitRateDisabled;
     OCTToxAVVideoBitRate videoBitRate = (enableVideo) ? kDefaultVideoBitRate : kOCTToxAVVideoBitRateDisable;
```
dvor, line 131: Why does `video` constant has `k` preffix but `audio` hasn't?

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -335,6 +362,43 @@ - (void)checkForCurrentActiveCallAndPause
...
+
+        return [self.toxAV setVideoBitRate:bitrate force:NO forFriend:friend.friendNumber error:error];
+    }
+    else {
+        // TO DO: Group Calls
+        return NO;
+    }
+}
+
+- (void)start:(BOOL)start enginesForCall:(OCTCall *)call
```
dvor: Same with this method naming. `startEngines:forCall:` would be better.

dvor: Even `startEnginesAndTimer:forCall:`

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -327,7 +354,7 @@ - (void)updateCall:(OCTCall *)call withState:(OCTToxAVCallState)state pausedStat
 
 - (void)checkForCurrentActiveCallAndPause
 {
-    if (! [self.audioEngine isAudioRunning:nil]) {
+    if (! [self.audioEngine isAudioRunning:nil] && ! [self.videoEngine isVideoSessionRunning]) {
```
dvor: Hm, if video is disabled then
- `audioEngine` would be running
- `videoEngine` wouldn't be running
- call wouldn't be put on pause because of `video`

Classes/Private/Manager/Video/OCTVideoEngine.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Indicates if the video session is running.
+ * @return YES if running, NO otherwise.
+ */
+- (BOOL)isVideoSessionRunning;
+
+/**
+ * Current video feed.
+ */
+- (UIView *)videoFeed;
```
dvor: I'm wondering maybe we want to provide CALayer for videoFeed too? AFAIK this would make it easier to port objcTox on Mac, a there are CALayer's too.

dvor:
This looks good so far. I'll take more detailed look at video engine later today. :-)

Classes/Private/Manager/Video/OCTVideoEngine.h
Chuongv: Hmm I'll have to look into that. Not sure if there's a way to grab CALayer from GLKView instead.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 131: There are no enums for video bitrates yet. Where as there are for audio.

Classes/Private/Manager/Video/OCTVideoEngine.h
dvor: view.layer? :-)

```
Dmytro
```

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 131: Once I figure out good video bitrates to use, I'll create the enums. I don't even know if 400 is a good default number? Any ideas?

Classes/Private/Manager/Video/OCTVideoEngine.h
Chuongv: Oh haha, didn't know it was that simple.

dvor: However I'm not sure if it is good idea to get later from GLKView. Need to read docs and investigate this.

Maybe we can return UIView instead of layer in other view instead?

```
Dmytro
```

Chuongv:
Hmm tests are still failing... weird. I can't reproduce the failure on either iphone 4s or iphone 6 sim.

Chuongv:
Well that's embarassing, now it passes :smile:

Classes/Private/Manager/Objects/OCTCall.m
```diff
@@ -29,4 +29,9 @@ - (NSDate *)onHoldDate
     return [NSDate dateWithTimeIntervalSince1970:self.onHoldStartInterval];
 }
 
+- (BOOL)isPaused
+{
+    return ! (self.pausedStatus == OCTCallPausedStatusNone);
```
dvor: `return (self.pausedStatus != OCTCallPausedStatusNone);` :-)

dvor: This method doesn't use any private properties of OCTCall. I think that you can move it to separate category file `OCTCall+Private.m`

dvor: And even better it would be to rename it to something like `OCTCall+Utilities`

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +99,44 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
     return nil;
 }
 
+- (BOOL)enableVideoSending:(BOOL)enable forCall:(OCTCall *)call error:(NSError **)error
+{
+    OCTToxAVVideoBitRate bitrate = (enable) ? kDefaultVideoBitRate : kOCTToxAVVideoBitRateDisable;
+    BOOL success = [self setVideoBitrate:bitrate forCall:call error:error];
```
dvor, line 105: Is there any need to set video bitrate to 0 when video is disabled? You are stopping videoEngine anyway

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
dvor, line 131: No ideas, leave it for now like that. :-)

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -87,11 +99,44 @@ - (OCTCall *)callToChat:(OCTChat *)chat enableAudio:(BOOL)enableAudio enableVide
...
+        return NO;
+    }
+
+    if (enable && (! [call isPaused])) {
+        OCTFriend *friend = [call.chat.friends firstObject];
+        self.videoEngine.friendNumber = friend.friendNumber;
+        [self.videoEngine startSendingVideo];
+    }
+    else {
+        [self.videoEngine stopSendingVideo];
```
dvor, line 117: There is some code duplication in this method and `startEngineAndTimer:forCall:`. I'm thinking if we can refactor all this code. Can we use single method `updateAudioEngines` that will
- get active call
- check if audio is required for call
  - audio required, check if audioEngine is running and start it if needed
  - audio not required, check if audioEngine is off and stop it if needed
- check if video is required for call
  - -||-
  - -||-

In this case all over the submanager you'll need only to update active call and call `updateAudioEngines` method

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 105: I think when you set it to 0, toxav notifies the friend that you stopped sending video. That way the other user can update the UI appropriately.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 117: Didn't push changes for this yet. Will start looking over this right now.

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
dvor, line 105: Okay, I see

Classes/Private/Manager/Video/OCTVideoEngine.h
```diff
@@ -0,0 +1,100 @@
...
+ * Start sending video data.
+ * This will turn on processIncomingVideo to YES
+ */
+- (void)startSendingVideo;
+
+/**
+ * Stop sending video data.
+ * This will turn off processIncomingVideo to NO
+ */
+- (void)stopSendingVideo;
```
dvor, line 50: It would nice to know if video is being send. Maybe change `{start,stop}SendingVideo` to `sendVideo` property?

dvor, line 50: Oh, there is `isVideoSessionRunning` method. I'm okay with 3 methods for that, but could you rename `isVideoSessionRunning` to `isSendingVideo` to be more consistent?

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+
+@property (nonatomic, assign) NSUInteger sizeOfChromaPlanes;
+
+@end
+
+@implementation OCTVideoEngine
+
+- (instancetype)init
+{
+    DDLogVerbose(@"%@: init", self);
```
dvor: `self` isn't initialized here, please move this log below `if (! self)` check

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+    self = [super init];
+    if (! self) {
+        return nil;
+    }
+
+    _captureSession = [AVCaptureSession new];
+    _captureSession.sessionPreset = AVCaptureSessionPreset1280x720;
```
dvor, line 45: For the future - this should be configurable. It would be nice to change quality depending of bandwidth.

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+    return self;
+}
+
+#pragma mark - Public
+
+- (BOOL)setupWithError:(NSError **)error
+{
+    DDLogVerbose(@"%@: setupWithError", self);
+    AVCaptureDevice *videoCaptureDevice = [self frontCamera];
+    AVCaptureDeviceInput *videoInput = [AVCaptureDeviceInput deviceInputWithDevice:videoCaptureDevice error:error];
```
dvor, line 58: For the future - we would need ability to switch between front and back cameras

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+    if (! videoInput) {
+        return NO;
+    }
+
+    [self.captureSession addInput:videoInput];
+
+    [self.dataOutput setAlwaysDiscardsLateVideoFrames:YES];
+    NSNumber *value  = [NSNumber numberWithUnsignedInt:kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange];
+    [self.dataOutput setVideoSettings:[NSDictionary dictionaryWithObject:value
+                                                                  forKey:(id)kCVPixelBufferPixelFormatTypeKey]];
```
dvor: More modern:

``` objc
self.dataOutput.alwaysDiscardsLateVideoFrames = YES;
self.dataOutput.videoSettings = @{
    kCVPixelBufferPixelFormatTypeKey : @(kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange),
};
```

or if cast is needed

``` objc
self.dataOutput.alwaysDiscardsLateVideoFrames = YES;
self.dataOutput.videoSettings = @{
    (NSString *)kCVPixelBufferPixelFormatTypeKey : @(kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange),
};
```

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+    return YES;
+}
+
+- (void)startSendingVideo
+{
+    DDLogVerbose(@"%@: startSendingVideo", self);
+    self.processIncomingVideo = YES;
+    if ([self isVideoSessionRunning]) {
+        return;
+    }
```
dvor: For thread safety it is better to perform this check on `processingQueue`. Same for `stopSendingVideo` and `isVideoSessionRunning` methods

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -0,0 +1,246 @@
...
+#import "OCTVideoView.h"
+#import "DDLog.h"
+
+#undef LOG_LEVEL_DEF
+#define LOG_LEVEL_DEF LOG_LEVEL_VERBOSE
+
+@import AVFoundation;
+
+static uint8_t *reusableUChromaPlane;
+static uint8_t *reusableVChromaPlane;
```
dvor: Why this need to be static instead of engine's property?

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv, line 117: Hmm I'm going to have to revisit this later. Tried to change it but then I had some trouble figuring out how to mock audioEngine since I needed to stub `[self.audioEngine isAudioRunning:nil]` to both YES and NO in one method.

Classes/Private/Manager/Video/OCTVideoEngine.m
Chuongv: If I were to put the `isVideoSessionRunning` on the processingQueue, that would mean I have to provide a completion block to the user correct?

dvor:
```
__block BOOL isRunning = NO;

dispatch_sync_bla_bla {
    isRunning = self.captureSession.isRunning;
};

return isRunning;
```

dvor: Or if captureSession is threadsafe there is no need for that. However I'm not sure about it, you can check documentation

Chuongv: Yea there is no mention about it being a blocking call so I think we are safe there.

PR #84, opened 2015-07-31
Several fixes to clean up warnings

Chuongv:
Cleaned up warnings

### DIFF
--- a/Classes/Private/Manager/Audio/OCTAudioEngine.m
+++ b/Classes/Private/Manager/Audio/OCTAudioEngine.m
@@ -308,13 +308,13 @@ OSStatus outputRenderCallBack(void *inRefCon,
 {
     OCTAudioEngine *myEngine = (__bridge OCTAudioEngine *)inRefCon;
 
-    UInt32 targetBufferSize = ioData->mBuffers[0].mDataByteSize;
+    int32_t targetBufferSize = ioData->mBuffers[0].mDataByteSize;
     SInt16 *targetBuffer = (SInt16 *)ioData->mBuffers[0].mData;
 
     int32_t availableBytes;
     SInt16 *buffer = TPCircularBufferTail(&myEngine->_outputBuffer, &availableBytes);
 
-    if (availableBytes < (int32_t)targetBufferSize) {
+    if (availableBytes < targetBufferSize) {
         memset(targetBuffer, 0, targetBufferSize);
         return noErr;
     }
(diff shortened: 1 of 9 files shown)


Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -257,7 +257,11 @@ - (void)pause:(BOOL)pause controlSentToCall:(OCTCall *)call
             callToUpdate.pausedStatus |= OCTCallPausedStatusByUser;
         }
         else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wassign-enum"
             callToUpdate.pausedStatus &= ~OCTCallPausedStatusByUser;
+#pragma clang diagnostic pop
+
```
Chuongv: Tried type casting, but the warnings still show up here. Thus resorting to this instead.

Classes/Private/Manager/Audio/OCTAudioEngine.m
```diff
@@ -314,7 +314,7 @@ OSStatus outputRenderCallBack(void *inRefCon,
     int32_t availableBytes;
     SInt16 *buffer = TPCircularBufferTail(&myEngine->_outputBuffer, &availableBytes);
 
-    if (availableBytes < targetBufferSize) {
+    if (availableBytes < (int32_t)targetBufferSize) {
```
dvor: I think it would be better to change `targetBufferSize` type ot `int32_t`

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
dvor: That's bad. I've turned off `CLANG_WARN_ASSIGN_ENUM` flag so this won't be an issue anymore. You can rebase on top of `audio` branch and remove this code.

dvor:
Please note that there are some more errors when building for real device.

Also there are lots of analyzer issues. Will you fix them as well (here or in separate pull request)?

Chuongv:
> Please note that there are some more errors when building for real device.

Oh okay I see it now. I was only building for simulator before. :+1:

dvor:
@Chuongv  just in case if you missed - there are still two unresolved comments. :-)

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
dvor: Also don't forget to remove this stuff in other place

Chuongv:
> @Chuongv just in case if you missed - there are still two unresolved comments. :-)

Thanks, forgot about the other two. Let me know if I missed anything else :smile:

PR #91, opened 2015-08-06
Video updates

Chuongv:
I am pretty sure the sending of video is correct now. The playback.. however I am having trouble with. Not sure if I understand strides correctly but here's what echobot is showing. But this could also be how echobot is striping the strides and sending it back to.

Top video is from the echobot, bottom is the preview

Image of echoBot preview

### DIFF
--- a/Classes/Private/Manager/Video/OCTPixelBufferPool.m
+++ b/Classes/Private/Manager/Video/OCTPixelBufferPool.m
@@ -1,0 +1,106 @@
+//
+//  OCTPixelBufferPool.m
+//  objcTox
+//
+//  Created by Chuong Vu on 8/5/15.
+//  Copyright (c) 2015 dvor. All rights reserved.
+//
+
+#import "OCTPixelBufferPool.h"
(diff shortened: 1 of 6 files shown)


Chuongv:
Fix the green color. Calculated wrong the interleaving of UV Planes. Need to see if I can up the resolution, since it's only sending at 192x144. Toxav wasn't accepting higher before
Image of Video

dvor:
Hurray, video! :tada:

Good job, will try it later today. :+1:

Classes/Private/Manager/Video/OCTPixelBufferPool.m
```diff
@@ -0,0 +1,106 @@
...
+
+@end
+
+@implementation OCTPixelBufferPool
+
+#pragma mark - Lifecycle
+
+- (instancetype)initWithFormat:(OSType)format;
+{
+    self = [super self];
```
dvor, line 30:
```
if (! self) {
    return nil;
}
```

Issue #92, opened 2015-08-06
Enable video in the middle of an audio call crashes objcTox


Chuongv:
Seems to have a long time enabling the video and it locks up causing the OCTAudioEngine to crash as well.

PR #93, opened 2015-08-06
Use completion blocks for video and preview view retrieval

Chuongv:
Fixes https://github.com/Antidote-for-Tox/objcTox/issues/92

### DIFF
--- a/Classes/Private/Manager/Video/OCTVideoEngine.h
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.h
@@ -56,16 +56,16 @@
 - (BOOL)isSendingVideo;
 
 /**
- * Current video feed.
+ * Generate a UIView with the current incoming video feed.
  */
-- (UIView *)videoFeedWithRect:(CGRect)rect;
+- (void)getVideoFeed:(void (^)(UIView *videoFeed))completionBlock;
 
 /**
  * Layer of the preview video.
  * @return CALayer that can be added to a UIView. Nil if
  * OCTVideoEngine is not sending any video.
  */
-- (CALayer *)videoCallPreview;
+- (void)getVideoCallPreview:(void (^)(CALayer *layer))completionBlock;
 
 /**
  * Provide video frames to video engine to process.
--- a/Classes/Private/Manager/Video/OCTVideoEngine.h
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.h
@@ -29,15 +29,6 @@
 - (BOOL)setupWithError:(NSError **)error;
 
 /**
- * Process incoming video frames.
- * Set to YES to process video, otherwise NO
- * to ignore the incoming frames in receiveVideoFrameWithWidth...
- * Will be set to YES automatically when calling startSendingVideo
- * Will be set to NO after stopSendingVideo
- */
-@property (nonatomic, assign) BOOL processIncomingVideo;
-
-/**
  * Start sending video data.
  * This will turn on processIncomingVideo to YES
  */
@@ -58,7 +49,7 @@
 /**
  * Generate a UIView with the current incoming video feed.
  */
-- (void)getVideoFeed:(void (^)(UIView *videoFeed))completionBlock;
+- (UIView *)videoFeed;
 
 /**
  * Layer of the preview video.
--- a/Classes/Private/Manager/Video/OCTVideoEngine.h
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.h
@@ -53,8 +53,9 @@
 
 /**
  * Layer of the preview video.
- * @return CALayer that can be added to a UIView. Nil if
- * OCTVideoEngine is not sending any video.
+ * Layer will be nil if videoSession is not running.
+ * @param completionBlock Block responsible for using the layer. This
+ * must not be nil.
  */
 - (void)getVideoCallPreview:(void (^)(CALayer *layer))completionBlock;
 
--- a/Classes/Private/Manager/Video/OCTVideoEngine.m
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.m
@@ -128,7 +128,10 @@ static const OSType kPixelFormat = kCVPixelFormatType_420YpCbCr8BiPlanarVideoRan
     DDLogVerbose(@"%@: videoCallPreview", self);
     dispatch_async(self.processingQueue, ^{
         AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
-        completionBlock(previewLayer);
+
+        dispatch_async(dispatch_get_main_queue(), ^{
+            completionBlock(previewLayer);
+        });
     });
 }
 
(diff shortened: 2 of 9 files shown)


Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -125,22 +125,24 @@ - (BOOL)isSendingVideo
...
-        self.videoView = [[OCTVideoView alloc] initWithFrame:rect];
-    }
-
-    return self.videoView;
+    DDLogVerbose(@"%@: getVideoFeed", self);
+    dispatch_async(self.processingQueue, ^{
+        if (! self.videoView) {
+            self.videoView = [[OCTVideoView alloc] initWithFrame:CGRectZero];
+        }
+        completionBlock(self.videoView);
```
dvor: I think it would be better to return feed and preview on main queue. So you can do dispatch_async once again. This way there will be no need for changing queue in client's code.

Also I'm wondering, wouldn't it be better to pefrorm sync operation and return preview/feed? Or it may block?

```
- (UIView *)videoFeed
{
    __block UIView *feed;

    dispatch_sync(...{
        feed = ...
    }
    return feed;
}
```

dvor: And one more thing. Creating view on background queue looks wrong to me, as UIKit isn't thread safe and _should_ be used only on main queue.

Chuongv: Hmm I'll try the above and see if it makes a difference. Right now this is what Antidote looks like when setting up the video screens. Which yea, looks kind of ugly?

```
    UIView *loadingView = [UIView new];
    loadingView.backgroundColor = [UIColor grayColor];
    controller.videoView = loadingView;

    __weak ActiveCallViewController *weakActiveVC = controller;
    [self.manager getVideoFeed:^(UIView *videoFeed) {
        ActiveCallViewController *strongActiveVC = weakActiveVC;

        dispatch_async(dispatch_get_main_queue(), ^{
            strongActiveVC.videoView = videoFeed;
        });
    }];

    [self.manager getVideoCallPreview:^(CALayer *layer) {
        ActiveCallViewController *strongActiveVC = weakActiveVC;
        dispatch_async(dispatch_get_main_queue(), ^{
            strongActiveVC.previewLayer = layer;
        });
    }];
```

Chuongv: Yea doing the below

```
- (UIView *)videoFeed
{
    __block UIView *feed;

    dispatch_sync(...{
        feed = ...
    }
    return feed;
}
```

Blocks completely on the main queue, and it still takes a long time if it's on the processing queue.

dvor: Can you profile what's taking long time? Maybe creating EAGLContext?

I really don't think it is good idea to create view _not_ on main queue. If context is taking time than maybe we can create it on start? If that won't work then we can go

```
- (UIView *)videoFeed
{
    __block UIView *feed;

    dispatch_async(...{
        // do stuff taking long time

        dispatch_async(main_queue) {
            // create videoView
            completionBlock(self.videoView);
        }
    }
    return feed;
}
```

dvor: Also as an option, we can
- create view on main thread and return immediately
- create context asynchronously 
- while there is no context draw just black view in `drawRect:`

Chuongv: Yea just confirmed it is the `EAGLContext *eaglContext = [[EAGLContext alloc] initWithAPI:kEAGLRenderingAPIOpenGLES2];` that is taking a long time. Will see if I can just put that on the processing queue instead.

dvor: Do we need to create context on `processingQueue`?
- if yes, you can pass it to VideoView
- if no, you can simply create queue on VideoView for creating context _or_ you can use system background queue for that

Chuongv: Ah okay good idea.

Chuongv: Ah okay, was able to get the eaglContext on the background queue. Tried putting the dispatch_async on the main queue for preview screen but that still locks it up for a long time. I think we have to keep `AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];` on the processing queue.

Chuongv:
I also removed the `processIncomingVideo` variable. This time it will check if the videoView is present, if it's not present, don't process. The client will be responsible for holding a strong reference to the videoView. If client removes the view, the videoEngine will stop processing incoming video.

Chuongv:
Only problem still is that I can't get the view to be black while the eaglContext is loading. drawRect won't actually run unless the context is there. Also setting the self.backgroundColor to black won't do it either.

dvor:
> Only problem still is that I can't get the view to be black while the eaglContext is loading. drawRect won't actually run unless the context is there. Also setting the self.backgroundColor to black won't do it either.

What color it is if not black? Maybe it is okay to leave this color?

May this be because of `enableSetNeedsDisplay == NO`? You can try to set it to YES while there is no context, then set back to NO.

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -125,22 +123,28 @@ - (BOOL)isSendingVideo
...
+        OCTVideoView *feed = [[OCTVideoView alloc] initWithFrame:CGRectZero];
+        self.videoView = feed;
 
-    return self.videoView;
+        return feed;
+    }
+    else {
+        return self.videoView;
+    }
 }
```
dvor, line 151: This looks nicer to me

```
- (UIView *)videoFeed;
{
    DDLogVerbose(@"%@: videoFeed", self);

    OCTVideoView *feed = self.videoView;

    if (! feed) {
        feed = [[OCTVideoView alloc] initWithFrame:CGRectZero];
        self.videoView = feed;
    }

    return feed;
}
```

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -125,22 +123,28 @@ - (BOOL)isSendingVideo
...
 
-- (CALayer *)videoCallPreview
+- (void)getVideoCallPreview:(void (^)(CALayer *))completionBlock
 {
     DDLogVerbose(@"%@: videoCallPreview", self);
-    AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
-
-    return previewLayer;
+    dispatch_async(self.processingQueue, ^{
+        AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
```
dvor, line 131: Should layer be created on `processingQueue`?
- yes - can we dispatch_async again after creating and return it on main queue?
- no - no need for async?

Chuongv:
> What color it is if not black? Maybe it is okay to leave this color?

It's just transparent, clear view. It's not too bad since it's only there for less than a second.

dvor:
Hm, I think it is okay to leave it transparent for now.

Classes/Private/Manager/Video/OCTVideoEngine.h
```diff
@@ -56,16 +47,16 @@
...
-- (UIView *)videoFeedWithRect:(CGRect)rect;
+- (UIView *)videoFeed;
 
 /**
  * Layer of the preview video.
  * @return CALayer that can be added to a UIView. Nil if
  * OCTVideoEngine is not sending any video.
  */
-- (CALayer *)videoCallPreview;
+- (void)getVideoCallPreview:(void (^)(CALayer *layer))completionBlock;
```
dvor, line 60: Can you please add documentation that completionBlock is a required parameter? You can also add an NSAssert for that

Chuongv, line 60: Sure thing.

PR #95, opened 2015-08-07
Stop video session when call is ended by friend

Chuongv:
Stop video session when call is ended by friend


Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -428,6 +428,7 @@ - (void)toxAV:(OCTToxAV *)toxAV callStateChanged:(OCTToxAVCallState)state friend
 
         if ((self.audioEngine.friendNumber == friendNumber) && [self.audioEngine isAudioRunning:nil]) {
             [self.audioEngine stopAudioFlow:nil];
+            [self.videoEngine stopSendingVideo];
```
dvor, line 434: Can we have situation when audioEngine is not running, but videoEngine is? In that case it won't stop.

Chuongv, line 434: Currently not but I guess it would be better to put the stop video engine in a separate line to guard against it in the future.

`if ((self.videoEngine.friendNumber == friendNumber) && [self. self.videoEngine isRunning:nil])`

Issue #94, opened 2015-08-07
Account for stride when capturing video


Chuongv:
In `(void)captureOutput:(AVCaptureOutput *)captureOutput didOutputSampleBuffer:(CMSampleBufferRef)sampleBuffer fromConnection:(AVCaptureConnection *)connection` of `OCTVideoEngine` we are taking in the width and height of the planar data without accounting for the possibility of stride. See `CVPixelBufferGetBytesPerRowOfPlane`

The video sending looks fine as of right now, however if there is ever stride data, we will not be sending over the correct data.

PR #96, opened 2015-08-08
Account for strides

Chuongv:
This should take care of this issue https://github.com/Antidote-for-Tox/objcTox/issues/94

If someone sends a bottom up image, `OCTVideoEngine` should be able to process it.


Chuongv:
Closing need to rename one of the commits

PR #104, opened 2015-08-13
File transfers

stal888:
Working through the list from the other PR:
- [x] DDLog macro isn't working (there is no global log level for now, you have to define it manually in each file using DDLog (see example in OCTTox)).
- [x] Generics aren't available yet, so they don't work. Please remove them. I don't think that we should migrate to Xcode 7 as it wasn't released yet.
- [x] Some includes are missing.
- [x] Existing tests don't compile too (OCTTox callback was changed).
- [ ] There are no tests. Code should be covered with tests as much as possible.
- [ ] File transfers aren't implemented in demo project.
- [x] As database changed, migration for Realm should be done.


stal888:
@dvor Where is migration being performed in the current master?

dvor:
It isn't yet, we have to add it. I'll do basic stuff today and merge into file-transfer.

stal888:
Are the nullability things OK? Travis is failing on them.

dvor:
Yes, they should be okay. I've added `osx_image: xcode6.4` to travis, let's see if it will build
https://travis-ci.org/Antidote-for-Tox/objcTox/builds/75495112

dvor:
Hey @stal888,

I've fixed Travis and added [Realm migration](https://github.com/Antidote-for-Tox/objcTox/commit/e136b68f067061a11996357dc2af5646d4f88f68). As you just add new fields to model I think there is no need for you to do anything about it.

All changes are merged to `file-transfer`, you can merge it into your branch.

Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -21,7 +21,7 @@
 #undef LOG_LEVEL_DEF
 #define LOG_LEVEL_DEF LOG_LEVEL_VERBOSE
 
-static const uint64_t kCurrentSchemeVersion = 1;
+static const uint64_t kCurrentSchemeVersion = 7;
```
dvor: I think we have different views on scheme version. What I was thinking about is to increment this number
- only for releases of objcTox
- only if Realm model has changed

So the main idea is to give "user" of objcTox (Antidote or Nao) easy way to upgrade on new version without loosing all data. On the other hand, while developing objcTox you don't want to increase number any time you change model. This should be only once per release.

So, the flow should be following. When you are changing model
- only if it is first change in release, add new version
- check if you should add something in migration block.

Please change version back to 1.

Classes/Private/Manager/Database/OCTRealmManager.m
```diff
@@ -268,6 +301,17 @@ - (OCTMessageAbstract *)addMessageWithText:(NSString *)text
     return messageAbstract;
 }
 
+- (void)noteMessageFileChanged:(OCTMessageFile *)messageFile
+{
+    OCTMessageAbstract *abstractMessage = [messageFile linkingObjectsOfClass:[OCTMessageAbstract className] forProperty:@"messageFile"].firstObject;
```
dvor: I think it would be better to do `forProperty:NSStringFromSelector(@selector(messageFile))` to get error if property will change.

Classes/Private/Manager/OCTManager.m
```diff
@@ -74,12 +74,7 @@ - (instancetype)initWithConfiguration:(OCTManagerConfiguration *)configuration
         savedData = [NSData dataWithContentsOfFile:savedDataPath];
     }
 
-    _tox = [[OCTTox alloc] initWithOptions:configuration.options savedData:savedData error:error];
-
-    if (! _tox) {
-        return nil;
-    }
-
+    _tox = [[OCTTox alloc] initWithOptions:configuration.options savedData:savedData error:nil];
```
dvor: It looks like bad resolve of merge conflict. It should be vice versa, the red code is what we need.

Classes/Private/Manager/OCTManager.m
```diff
@@ -93,10 +88,34 @@ - (instancetype)initWithConfiguration:(OCTManagerConfiguration *)configuration
...
+    _files = files;
+
+    OCTSubmanagerAvatars *avatars = [OCTSubmanagerAvatars new];
+    avatars.dataSource = self;
+    _avatars = avatars;
+
+    OCTSubmanagerObjects *objects = [OCTSubmanagerObjects new];
+    objects.dataSource = self;
+    _objects = objects;
+
```
dvor: Please remove this block of code, submanagers are being created above with one-line method.

Classes/Private/Manager/OCTManager.m
```diff
@@ -156,6 +175,11 @@ - (OCTRealmManager *)managerGetRealmManager
     return self.configuration.fileStorage;
 }
 
+- (dispatch_queue_t)managerGetToxQueue
```
dvor, line 159: This method is unused.

Classes/Private/Manager/OCTManager.m
```diff
@@ -243,6 +266,21 @@ - (void)saveTox
...
+    [self.files sendProgressNotificationsNow];
+}
+
+- (void)tox:(OCTTox *)tox friendConnectionStatusChanged:(OCTToxConnectionStatus)status friendNumber:(OCTToxFriendNumber)friendNumber
+{
+    // the rightful owner of this callback
+    [self.friends tox:tox friendConnectionStatusChanged:status friendNumber:friendNumber];
+    // needed for file resuming machinery
+    [self.files tox:tox friendConnectionStatusChanged:status friendNumber:friendNumber];
+}
```
dvor, line 258: Maybe instead of breaking "forwarding" logic it would be better to implement some broadcast system that will allow submanagers to communicate? Any submanager will be able to broadcast some message and all other submanagers would be able to subscribe to it. So the logic would be
- submanager receives some event (friend connection status in our case)
- it send broadcast message to some broadcast center for all submanagers
- any interested submanager can subscribe to it (files submanager in our case).

However it may not be part of this issue, so we can leave it for now. However if you are interested you can implement it.

Classes/Private/Manager/OCTManager.m
```diff
@@ -243,6 +266,21 @@ - (void)saveTox
     }
 }
 
+#pragma mark - OCTToxDelegate
+
+- (void)toxDidIterate:(OCTTox *)tox
```
dvor: Do we really want to update file progress every 50ms? Maybe it would be better to have some timer on OCTSubmanagerFiles with more suitable update interval? This way there will be no need to expose didIterate method from OCTTox.

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+#import "OCTSubmanagerFiles.h"
+#import "OCTActiveFile.h"
+
+#define AVERAGE_SECONDS 10
+
+@class OCTMessageFile;
+
+@interface OCTActiveFile ()
+
+@property (weak)      OCTSubmanagerFiles *fileManager;
```
dvor: To be more consistent with the rest of the project, can you please add property attributes to all properties all over you change (except for Realm objects - Realm ignores those attributes)? Format

``` objc
(strong/copy/weak/assign, atomic/nonatomic, [readwrite, readonly])
```

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+// Observation: I'd rather not hold on to a Realm object. Who knows what could
+//              happen.
```
dvor: > Observation: I'd rather not hold on to a Realm object. Who knows what could happen.

Maybe it would be better to store `uniqueIdentifier` of object and get it with `[OCTRealmManager objectWithUniqueIdentifier:]` when needed?

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+//  objcTox
+//
+//  Created by stal on 9/7/2015.
+//  Copyright © 2015 Zodiac Labs. All rights reserved.
+//
+
+#import "OCTSubmanagerFiles.h"
+#import "OCTActiveFile.h"
+
+#define AVERAGE_SECONDS 10
```
dvor: Please use constants instead of defined

``` objc
static const NSUInteger kOCTActiveFileAverageSeconds = 10;
```

Also this constant can be moved to `.m` file, in this case naming can be simpler

``` objc
static const NSUInteger kAverageSeconds = 10;
```

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+@property             OCTToxFriendNumber friendNumber;
+@property             OCTToxFileSize fileSize;
+@property (readwrite) OCTToxFileSize bytesMoved;
+@property             BOOL isConduitOpen;
+
+@property             time_t lastCountedTime;
+@property             OCTToxFileSize     *transferRateCounters;
+@property             long rollingIndex;
+
+@property (strong)    OCTFileNotificationBlock notificationBlock;
```
dvor: Blocks should be `copy`

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+/* Helpful if a bit unclean. */
+- (id<OCTFileConduit>)conduit;
+
+- (void)sendProgressUpdateNow;
+- (void)control:(OCTToxFileControl)ctl;
+- (void)interrupt;
+
+@end
+
+@interface OCTActiveOutboundFile : OCTActiveFile
```
dvor: Messages use "outgoing" and "incoming" words. Maybe be consistent and use `OCTActiveOutgoingFile`, `OCTActiveIncomingFile`?

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+#undef LOG_LEVEL_DEF
+#define LOG_LEVEL_DEF LOG_LEVEL_DEBUG
+
+/* these are annoying, and because we get all file numbers from toxcore,
+ * casting will never truncate them because they were never longs to begin
+ * with. */
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wshorten-64-to-32"
+
+static time_t OCTGetSystemUptime(void)
```
dvor: Just wondering, why not to use `[NSDate date]` instead?

Classes/Public/Manager/OCTManagerConstants.h
```diff
@@ -13,31 +13,76 @@ typedef NS_ENUM(NSUInteger, OCTFetchRequestType) {
...
+    /**
+     * Unimplemented.
+     */
+    OCTFileUsageInlineVideo,
+};
+
+/* note: signed type is being used because of Realm. */
+typedef NS_OPTIONS(NSInteger, OCTPauseFlags) {
+    OCTPauseFlagsSelf = 1,
+    OCTPauseFlagsOther = 1 << 1,
```
dvor: `OCTPauseFlagsOther` sounds like "other" flag, not "other" person. Maybe rename it to `OCTPauseFlagsFriend`?

Classes/Public/Manager/OCTManagerConstants.h
```diff
@@ -13,31 +13,76 @@ typedef NS_ENUM(NSUInteger, OCTFetchRequestType) {
...
+
+        /* These are for convenience. */
+        OCTPauseFlagsNobody = 0,
+        OCTPauseFlagsBoth = 3,
```
dvor: `OCTPauseFlagsBoth = OCTPauseFlagsSelf | OCTPauseFlagsOther,`

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+            DDLogDebug(@"_control: obeying cancel message from remote.");
+
+            dispatch_async(self.fileManager.queue, ^{
+                [self closeConduitIfNeeded];
+
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    [self.fileManager removeFile:self];
+                });
+            });
+            [self markFileAsCancelled:self.fileMessage];
```
dvor: You can simply call `[self interrupt]` here

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+    self.lastCountedTime = now;
+
+    NSAssert(delta >= 0, @"Detected a temporal anomaly. objcTox currently does not support Tox FTL, nor phone-microwave-"
+             "based time travel. Please file a bug.");
+
+    if (delta != 0) {
+        [self incrementRollingIndex:delta];
+    }
+
+    self.transferRateCounters[self.rollingIndex] += size;
```
dvor, line 127: Shouldn't this be in `@synchronized` block?

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,58 @@
...
+//              happen.
+@property (strong)    OCTMessageFile     *fileMessage;
+@property             OCTToxFriendNumber friendNumber;
+@property             OCTToxFileSize fileSize;
+@property (readwrite) OCTToxFileSize bytesMoved;
+@property             BOOL isConduitOpen;
+
+@property             time_t lastCountedTime;
+@property             OCTToxFileSize     *transferRateCounters;
+@property             long rollingIndex;
```
dvor: Lot's of properties are used in `OCTActiveFile.m` class only (for example `transferRateCounters`). Can you make them private by moving in `.m` file?

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+- (void)markFileAsCancelled:(OCTMessageFile *)file
+{
+    DDLogDebug(@"Marking file %@ as CANCELLED.", file);
+
+    [[self.fileManager.dataSource managerGetRealmManager] updateObject:file withBlock:^(OCTMessageFile *theObject) {
+        theObject.fileState = OCTMessageFileStateCanceled;
+        theObject.filePosition = 0;
+        theObject.restorationTag = nil;
+        theObject.fileTag = nil;
+        // theObject.fileNumber = 0;
```
dvor: No need for commented code. Please change or remove it.

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+
+- (void)markFileAsCompleted:(OCTMessageFile *)file withFinalDestination:(NSString *)fd
+{
+    DDLogDebug(@"Marking file %@ as COMPLETE.", file);
+
+    [[self.fileManager.dataSource managerGetRealmManager] updateObject:file withBlock:^(OCTMessageFile *theObject) {
+        theObject.fileState = OCTMessageFileStateReady;
+        theObject.restorationTag = nil;
+        theObject.fileTag = nil;
+        // theObject.fileNumber = 0;
```
dvor: Same for this one

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+    NSData *conduitData = nil;
+    if ([self.conduit conformsToProtocol:@protocol(NSCoding)]) {
+        conduitData = [NSKeyedArchiver archivedDataWithRootObject:self.conduit];
+    }
+
+    [[self.fileManager.dataSource managerGetRealmManager] updateObject:file withBlock:^(OCTMessageFile *theObject) {
+        theObject.fileState = OCTMessageFileStateLoading;
+        theObject.pauseFlags = OCTPauseFlagsNobody;
+        theObject.filePosition = self.bytesMoved;
+        theObject.restorationTag = conduitData;
```
dvor: Why you archive conduit when resuming file?

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,530 @@
...
+- (BOOL)resumeWithError:(NSError *__autoreleasing *)error
+{
+    BOOL ok = YES;
+    if (error) {
+        *error = nil;
+    }
+
+    if ((self.fileMessage.fileState != OCTMessageFileStatePaused) &&
+        (self.fileMessage.fileState != OCTMessageFileStateWaitingConfirmation) ) {
+        // TODO make OCTFileError a constant symbol
```
dvor: Will you do that? It would be add enum with error codes to `OCTManagerConstants`.

Classes/Private/Manager/Objects/OCTActiveFile.m
stal888: To properly count progress, a clock source that doesn't go back in time (the user can change the system date) is needed. The Mach clock is guaranteed to not do that.

(That said, the function name is wrong so I'll need to fix that)

Classes/Private/Manager/Objects/OCTActiveFile.m
stal888, line 127: It's probably okay. I think a mutex is too heavy-handed for a method that gets called so often. (a rough calculation `6 MB/s -> 6,000,000 bytes in ~1430 byte packets -> 4,195.8 message sends/s, yikes!`, and that's just this method)

Since this is only ever set from the Tox queue (serialized), the worst that could happen is the client gets a wrong estimated time value (and we exclude the current hot-second from the average) which will be wiped in the next progress update.

dvor:
Reviewed only half of pull request, will do other half later. :-)

Classes/Private/Manager/OCTManager.m
stal888, line 258: I'll do this in another branch. Avatars needs it too.

Classes/Private/Manager/Objects/OCTActiveFile.m
stal888: It's for the initial transition from WaitingConfirmation.

stal888:
There are more commits coming, this is just a heads-up so you don't review code only to have it removed a day later.

dvor:
Okay, waiting for more commits then.

stal888:
I failed to make the code any smaller so this is now ready to go, I guess.

Tests coming soon.

Classes/Private/Manager/OCTManager.m
```diff
@@ -74,7 +74,7 @@ - (instancetype)initWithConfiguration:(OCTManagerConfiguration *)configuration
         savedData = [NSData dataWithContentsOfFile:savedDataPath];
     }
 
-    _tox = [[OCTTox alloc] initWithOptions:configuration.options savedData:savedData error:error];
+    _tox = [[OCTTox alloc] initWithOptions:configuration.options savedData:savedData error:nil];
```
dvor: Please revert this line. :-)

Classes/Private/Manager/Objects/OCTActiveFile+Variants.h
```diff
@@ -0,0 +1,45 @@
...
+@class OCTMessageFile;
+
+@interface OCTActiveFile ()
+
+@property (weak, atomic)   OCTSubmanagerFiles *fileManager;
+// Realm identifier
+@property (copy, atomic)   NSString           *fileIdentifier;
+@property (atomic)         OCTToxFriendNumber friendNumber;
+@property (atomic)         OCTToxFileNumber fileNumber;
+@property (atomic)         OCTToxFileSize fileSize;
```
dvor: Can you please add attributes to this properties as well?

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+@property (atomic)    BOOL isConduitOpen;
+
+@property (atomic)    unsigned long lastCountedTime;
+@property (atomic)    unsigned long lastProgressUpdateTime;
+@property (atomic)    OCTToxFileSize *transferRateCounters;
+@property (atomic)    long rollingIndex;
+
+@property (copy)      OCTFileNotificationBlock notificationBlock;
+
+@property (atomic)    BOOL suppressNotifications;
```
dvor: Can you please add attributes to this properties as well? (2)

Classes/Private/Manager/Objects/OCTFileInput.m
```diff
@@ -0,0 +1,161 @@
...
+
+#undef LOG_LEVEL_DEF
+#define LOG_LEVEL_DEF LOG_LEVEL_DEBUG
+
+@interface OCTFileInput ()
+
+@property (copy)   NSString *path;
+@property (strong) NSFileHandle *readHandle;
+@property NSTimeInterval modifiedTime;
+@property OCTToxFileSize knownFileSize;
```
dvor: Can you please add attributes to this properties as well? (3)

Classes/Private/Manager/Objects/OCTFileOutput.m
```diff
@@ -0,0 +1,151 @@
...
+#import "OCTFileStorageProtocol.h"
+
+#undef LOG_LEVEL_DEF
+#define LOG_LEVEL_DEF LOG_LEVEL_DEBUG
+
+@interface OCTFileOutput ()
+
+@property (copy)   NSString *_finalPathName;
+@property (copy)   NSString *_temporaryPathName;
+@property (strong) NSFileHandle *_writeHandle;
```
dvor: Can you please add attributes to this properties as well? (4)

Classes/Public/Wrapper/OCTTox.h
```diff
@@ -17,6 +17,11 @@
 @property (weak, nonatomic) id<OCTToxDelegate> delegate;
 
 /**
+ * A GCD queue that can be used to run synchronously with the Tox loop.
+ */
+@property (strong) dispatch_queue_t queue;
```
dvor: Can you please add attributes to this property as well? (5)

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+    });
+}
+
+- (id<OCTFileConduit>)conduit
+{
+    @throw [NSException exceptionWithName:@"OCTFileException" reason:@"Only subclasses of OCTActiveFile can be used." userInfo:nil];
+    return nil;
+}
+
+- (BOOL)openConduitIfNeeded
```
dvor, line 220: I think that you can move `dispatch_sync` into `openConduitIfNeeded` method.

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+
+        self.isConduitOpen = ret;
+        return ret;
+    }
+    else {
+        return YES;
+    }
+}
+
+- (void)closeConduitIfNeeded
```
dvor: You can also move `dispatch_async` to this method and change it to 

``` objc
- (void)closeConduitIfNeededWithCompletion:(void (^)())completionBlock
```

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+    if (! self.isConduitOpen) {
+        return;
+    }
+    else {
+        [self.conduit transferWillBecomeInactive:self];
+        self.isConduitOpen = NO;
+    }
+}
+
+- (void)sendProgressUpdateNow
```
dvor, line 238: And `dispatch_async` can go here too as notification always has to be called on main thread.

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+    self.notificationBlock = blk;
+}
+
+#pragma mark - Realm update stuff
+
+- (void)markFileAsCancelled:(OCTActiveFile *)file
+{
+    DDLogDebug(@"Cancelling file %@", file);
+    self.suppressNotifications = YES;
+    [self.fileManager setState:OCTMessageFileStateCanceled forFile:file cleanInternals:YES andRunBlock:nil];
```
dvor, line 300: This refactoring is really nice :+1:

Classes/Private/Manager/Objects/OCTActiveFile.m
```diff
@@ -0,0 +1,504 @@
...
+}
+
+- (void)sendChunkForSize:(size_t)csize fromPosition:(OCTToxFileSize)p
+{
+    if (p != self.bytesMoved + 1) {
+        if ([self.sender respondsToSelector:@selector(moveToPosition:)]) {
+            [self.sender moveToPosition:p];
+        }
+        else {
+            DDLogWarn(@"OCTActiveInBoundFile WARNING: receiver %@ does not support seeking, but we received out of order file chunk for position %llu."
```
dvor: received -> sender; received -> sent

Classes/Private/Manager/Objects/OCTFileInput.m
```diff
@@ -0,0 +1,161 @@
...
+- (instancetype)initWithPath:(nonnull NSString *)path
+{
+    self = [super init];
+
+    if (self) {
+        self.path = path;
+        NSError *error = nil;
+        NSDictionary *attrs = [[NSFileManager defaultManager] attributesOfItemAtPath:[path stringByResolvingSymlinksInPath] error:&error];
+
+        if (error) {
```
dvor: You should check for `attrs` instead of `error`.

> Important Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning nil or NO, you should always check that the return value is nil or NO before attempting to do anything with the NSError object.

https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html

Classes/Private/Manager/Objects/OCTFileInput.m
```diff
@@ -0,0 +1,161 @@
...
+
+#pragma mark - NSCoding
+
+- (instancetype)initWithCoder:(nonnull NSCoder *)aDecoder
+{
+    self = [super init];
+
+    if (self) {
+        self.path = [aDecoder decodeObjectForKey:@"_filePath"];
+        self.modifiedTime = [aDecoder decodeDoubleForKey:@"modifiedTime"];
```
dvor: It would be better to make these strings static constants at the top of the file

dvor: Also why filePath has `_` prefix, but modifiedTime doesn't?

Classes/Private/Manager/Objects/OCTFileInput.m
```diff
@@ -0,0 +1,161 @@
...
+{
+    // nothing!
+}
+
+- (BOOL)canBeResumedNow
+{
+    NSError *error = nil;
+    NSDictionary *attrs = [[NSFileManager defaultManager] attributesOfItemAtPath:[self.path stringByResolvingSymlinksInPath] error:&error];
+
+    if (error) {
```
dvor: Please check for `attrs` instead of error

Classes/Private/Manager/Objects/OCTFileOutput.m
```diff
@@ -0,0 +1,151 @@
...
+
+// Call this when the user asks OCTSubmanagerFiles for a default saver.
+- (instancetype)initWithConfigurator:(id<OCTFileStorageProtocol>)cf
+{
+    self = [super init];
+
+    if (self) {
+        NSString *uuid = [NSUUID UUID].UUIDString;
+        self._finalPathName = [[cf pathForDownloadedFilesDirectory] stringByAppendingPathComponent:uuid];
+        self._temporaryPathName = [self._finalPathName stringByAppendingPathExtension:@"partial"];
```
dvor: Please move partial extension string to static constant.

Classes/Private/Manager/Objects/OCTFileOutput.m
```diff
@@ -0,0 +1,151 @@
...
+
+#pragma mark - <NSCoding>
+
+- (instancetype)initWithCoder:(nonnull NSCoder *)aDecoder
+{
+    self = [super init];
+
+    if (self) {
+        self._finalPathName = [aDecoder decodeObjectForKey:@"_finalPathName"];
+        self._temporaryPathName = [aDecoder decodeObjectForKey:@"_temporaryPathName"];
```
dvor: And these strings too

Classes/Private/Manager/Objects/OCTFileOutput.m
```diff
@@ -0,0 +1,151 @@
...
+    self._writeHandle = nil;
+    DDLogDebug(@"OCTFileOutput %@ closed.", self);
+}
+
+- (void)transferWillComplete:(nonnull OCTActiveFile *)file
+{
+    NSError *error = nil;
+    [[NSFileManager defaultManager] moveItemAtPath:self._temporaryPathName toPath:self._finalPathName error:&error];
+
+    if (error) {
```
dvor: Check for BOOL returned by `moveItemAtPath:toPath:error` instead of error

Classes/Private/Manager/Objects/OCTFileOutput.m
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#pragma mark - <OCTFileReceiving>
+
+- (void)writeBytes:(OCTToxFileSize)chunk_size fromBuffer:(nonnull const uint8_t *)buffer
+{
+    // possible 32bit bug: we are forcibly truncating chunk_size.
+    // The good news is, we're never going to encounter a chunk that big.
+    write(self._writeHandle.fileDescriptor, buffer, (size_t)chunk_size);
+    // [self._writeHandle writeData:[NSData dataWithBytesNoCopy:(uint8_t *)buffer length:(NSUInteger)chunk_size freeWhenDone:NO]];
```
dvor: No need for commented line

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -6,10 +6,93 @@
...
+
+static NSString *OCTSanitizeFilename(NSString *filename)
+{
+    // TODO: maybe get rid of nulls too
+    NSMutableString *mut = filename.mutableCopy;
+    [mut replaceOccurrencesOfString:@"/" withString:@"_" options:0 range:NSMakeRange(0, mut.length)];
+    if ([mut characterAtIndex:0] == '.') {
+        [mut replaceCharactersInRange:NSMakeRange(0, 1) withString:@"_"];
+    }
+    return mut;
```
dvor:
``` objc
return [mut copy];
```

dvor:
I almost done with reviewing that one, will finish in a few hours.

It looks really good, thanks for your work. :+1: It can be merged after few minor issues will be fixed. To speed everything up I think it would be better to push only fixes here, new stuff (including tests) can go to another pull request.

stal888:
All fixed

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+    NSParameterAssert(name);
+    NSParameterAssert(file);
+    NSParameterAssert(chat);
+
+    NSError *err = nil;
+
+    OCTFriend *f = chat.friends.firstObject;
+    OCTToxFileNumber n = [[self.dataSource managerGetTox] fileSendWithFriendNumber:f.friendNumber kind:OCTFileUsageToToxFileKind(type) fileSize:file.fileSize fileId:nil fileName:name error:&err];
+
+    if (err) {
```
dvor:
``` objc
if (n == kOCTToxFileNumberFailure) {
```

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+    [[self.dataSource managerGetRealmManager] updateObject:chat withBlock:^(OCTChat *theChat) {
+        theChat.lastMessage = msg;
+        theChat.lastActivityDateInterval = msg.dateInterval;
+    }];
+
+    if (msgout) {
+        *msgout = msg;
+    }
+
+    // [send resumeWithError:error];
```
dvor: No need for this

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+    }
+
+    OCTActiveIncomingFile *f = (OCTActiveIncomingFile *)[self realActiveFileForMessage:msg];
+    if (! [f isKindOfClass:[OCTActiveIncomingFile class]]) {
+        OCTExceptFileNotInbound();
+        return nil;
+    }
+
+    f.receiver = saver;
+    [f resumeWithError:error];
```
dvor: Hm, shouldn't we return `nil` if there was error while resuming?

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+    fmsg.pauseFlags = OCTPauseFlagsFriend;
+    fmsg.fileTag = [[self.dataSource managerGetTox] fileGetFileIdForFileNumber:n friendNumber:f.friendNumber error:nil];
+
+    msg.sender = nil;
+    msg.chat = chat;
+    msg.messageFile = fmsg;
+
+    OCTActiveOutgoingFile *send = (OCTActiveOutgoingFile *)[self createActiveFileForFriend:f message:fmsg provider:file isOutgoing:YES];
+    [self setActiveFile:send forFriendNumber:f.friendNumber fileNumber:n];
+    self.activeFiles[@(f.friendNumber)][@(n)] = send;
```
dvor: Use `setActiveFile:forFriendNumber:fileNumber:` method intead?

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+    NSError *error = nil;
+
+    OCTToxFileNumber n = [[self.dataSource managerGetTox] fileSendWithFriendNumber:f.friendNumber
+                                                                              kind:OCTFileUsageToToxFileKind(mf.fileUsage)
+                                                                          fileSize:sender.fileSize
+                                                                            fileId:mf.fileTag
+                                                                          fileName:mf.fileName
+                                                                             error:&error];
+
+    if (error) {
```
dvor:
``` objc
if (n == kOCTToxFileNumberFailure) {
```

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+        DDLogDebug(@"OCTSubmanagerFiles: receiver said NO for resuming file %@", msga);
+        return NO;
+    }
+
+    // now it can be resumed
+
+    NSError *error = nil;
+    [[self.dataSource managerGetTox] fileSeekForFileNumber:fileNumber friendNumber:msga.sender.friendNumber position:fileMsg.filePosition error:&error];
+
+    if (error) {
```
dvor:
``` objc
if (n == kOCTToxFileNumberFailure) {
```

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -25,7 +108,490 @@ - (instancetype)init
...
+        fmsg.fileNumber = fileNumber;
+        fmsg.fileSize = fileSize;
+        fmsg.fileName = OCTSanitizeFilename(fileName);
+        fmsg.fileUsage = OCTToxFileKindToFileUsage(kind);
+
+#ifdef OBJCTOX_SHOULD_BE_COMPATIBLE_WITH_UTOX_INLINE_IMAGES
+        if ([fileName isEqualToString:@"utox-inline.png"]) {
+            fmsg.fileUsage = OCTFileUsageInlinePhoto;
+        }
+#endif
```
dvor, line 575: Hm, maybe this stuff should be standardized in STS? :-)

Classes/Public/Manager/Objects/OCTActiveFile.h
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Current speed of the file transfer.
+ */
+@property (nonatomic, readonly) OCTToxFileSize bytesPerSecond;
+
+- (BOOL)pauseWithError:(NSError *__nullable *__nullable)error;
+- (BOOL)resumeWithError:(NSError *__nullable *__nullable)error;
+- (BOOL)cancelWithError:(NSError *__nullable *__nullable)error;
+
+- (void)beginReceivingLiveUpdatesWithBlock:(nullable OCTFileNotificationBlock)blk;
```
dvor: Maybe add any way of stopping receiving updates?

dvor:
I'm done with review. Please fix latest comments and we are good to merge.

Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
stal888, line 575: It should.

However, stickers are more important right now :smirk:.

Classes/Public/Manager/Objects/OCTActiveFile.h
stal888: The way to stop it is to pass nil. I'm just going to turn it into a property.

dvor:
Cool stuff

stal888:
Thanks for the review

Issue #106, opened 2015-08-13
Allow user to set videoOrientation on OCTVideoEngine


Chuongv:
See `AVCaptureVideoOrientation` this should be available on `OCTSubmanagerCalls`

Chuongv:
Auto rotation is taken care of automatically by `OCTVideoEngine`

PR #107, opened 2015-08-14
Add support for video rotation

Chuongv:
For https://github.com/Antidote-for-Tox/objcTox/issues/106 

Instead of allowing user to change it, objcTox will handle it automatically.

### DIFF
--- a/Classes/Private/Manager/Video/OCTVideoEngine.m
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.m
@@ -362,6 +362,7 @@ static const OSType kPixelFormat = kCVPixelFormatType_420YpCbCr8BiPlanarVideoRan
         case UIDeviceOrientationPortrait:
             orientation = AVCaptureVideoOrientationPortrait;
             break;
+        /* Landscapes are reversed, otherwise for some reason the video will be upside down */
         case UIDeviceOrientationLandscapeLeft:
             orientation = AVCaptureVideoOrientationLandscapeRight;
             break;
--- a/Classes/Private/Manager/Video/OCTVideoEngine.m
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.m
@@ -146,6 +146,8 @@ static const OSType kPixelFormat = kCVPixelFormatType_420YpCbCr8BiPlanarVideoRan
         dispatch_async(dispatch_get_main_queue(), ^{
             completionBlock(previewLayer);
         });
+
+        self.previewLayer = previewLayer;
     });
 }
 
--- a/Classes/Private/Manager/Video/OCTVideoEngine.m
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.m
@@ -65,6 +65,8 @@ static const OSType kPixelFormat = kCVPixelFormatType_420YpCbCr8BiPlanarVideoRan
     if (self.reusableVChromaPlane) {
         free(self.reusableVChromaPlane);
     }
+
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
 #pragma mark - Public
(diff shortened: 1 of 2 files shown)


Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -90,6 +91,10 @@ - (BOOL)setupWithError:(NSError **)error
     AVCaptureConnection *conn = [self.dataOutput connectionWithMediaType:AVMediaTypeVideo];
     conn.videoOrientation = AVCaptureVideoOrientationPortrait;
```
dvor: It would be better to call `orientationChanged` here instead of setting orientation, as it can be not Portrait by default.

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -130,10 +135,13 @@ - (void)getVideoCallPreview:(void (^)(CALayer *))completionBlock
     NSParameterAssert(completionBlock);
     DDLogVerbose(@"%@: videoCallPreview", self);
     dispatch_async(self.processingQueue, ^{
-        AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
+
+        if (! self.previewLayer) {
+            self.previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
```
dvor: Layer is weak and would be released immediately.

``` objc
dispatch_async(self.processingQueue, ^{
    AVCaptureVideoPreviewLayer *previewLayer = self.previewLayer;

    if (! previewLayer) {
        previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
        self.previewLayer = previewLayer;
    }

    dispatch_async(dispatch_get_main_queue(), ^{
        completionBlock(previewLayer);
    });
});
```

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -328,4 +336,39 @@ - (AVCaptureDevice *)frontCamera
     return nil;
 }
 
+- (void)registerOrientationNotification
+{
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(orientationChanged)
+                                                 name:UIDeviceOrientationDidChangeNotification
+                                               object:nil];
```
dvor, line 349: You have to add `removeObserver` method in dealloc or it will crash

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -130,7 +137,11 @@ - (void)getVideoCallPreview:(void (^)(CALayer *))completionBlock
     NSParameterAssert(completionBlock);
     DDLogVerbose(@"%@: videoCallPreview", self);
     dispatch_async(self.processingQueue, ^{
-        AVCaptureVideoPreviewLayer *previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
+        AVCaptureVideoPreviewLayer *previewLayer = self.previewLayer;
+
+        if (! self.previewLayer) {
+            previewLayer = [AVCaptureVideoPreviewLayer layerWithSession:self.captureSession];
```
dvor, line 143: Now `self.previewLayer` is never set

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -328,4 +341,39 @@ - (AVCaptureDevice *)frontCamera
...
+            orientation = AVCaptureVideoOrientationPortraitUpsideDown;
+            break;
+        case UIDeviceOrientationPortrait:
+            orientation = AVCaptureVideoOrientationPortrait;
+            break;
+        case UIDeviceOrientationLandscapeLeft:
+            orientation = AVCaptureVideoOrientationLandscapeRight;
+            break;
+        case UIDeviceOrientationLandscapeRight:
+            orientation = AVCaptureVideoOrientationLandscapeLeft;
```
dvor, line 370: Almost missed that. Why Left -> Right and Right -> Left? :grinning:

Chuongv, line 370: I don't know why, but if I don't do that the video will always be upside down when in landscape mode.

dvor, line 370: Can you please add a comment about that?

PR #108, opened 2015-08-15
Create video queue for OCTToxAV

Chuongv:
This alleviates this issue https://github.com/Antidote-for-Tox/Antidote/issues/111. Instead of crashing :100: % of the time, it is now only 10% . I think I should put the audio call backs on the same queue as well?


Chuongv:
Ah one of the tests failed. Will have to take a look tomorrow

Chuongv:
Closing, will open a different one that will better handle this.

PR #116, opened 2015-08-18
add some facilities for submanager communication

stal888:
Just a few things.
- Expose all submanagers as properties in OCTSubmanagerDataSource so they can talk directly
- Add a notification center to OCTSubmanagerDataSource so they can broadcast messages to interested submanagers. This isn't the defaultCenter so broadcasts are still framework-private.
- Add a notification post to tox:friendConnectionStatusChanged:friendNumber:.

P.S. can this be merged into file-transfers too?

### DIFF
--- a/objcToxTests/OCTSubmanagerFriendsTests.m
+++ b/objcToxTests/OCTSubmanagerFriendsTests.m
@@ -331,6 +331,19 @@ static NSString *const kMessage = @"kMessage";
     [self.submanager tox:self.tox friendConnectionStatusChanged:OCTToxConnectionStatusTCP friendNumber:kFriendNumber];
     XCTAssertEqual(friend.connectionStatus, OCTToxConnectionStatusTCP);
     XCTAssertTrue(friend.isConnected);
+
+    NSNotificationCenter *center = [[NSNotificationCenter alloc] init];
+    OCMStub([self.dataSource managerGetNotificationCenter]).andReturn(center);
+
+    XCTestExpectation *expect = [self expectationWithDescription:@""];
+    [center addObserverForName:kOCTFriendConnectionStatusChangeNotificationName object:nil queue:nil usingBlock:^(NSNotification *note) {
+        if ([note.object friendNumber] == kFriendNumber) {
+            [expect fulfill];
+        }
+    }];
+
+    [self.submanager tox:self.tox friendConnectionStatusChanged:OCTToxConnectionStatusUDP friendNumber:kFriendNumber];
+    [self waitForExpectationsWithTimeout:0.0 handler:nil];
 }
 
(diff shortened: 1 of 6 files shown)


dvor:
> P.S. can this be merged into file-transfers too?

Yes.

I'm wondering if we actually want to let submanagers talk directly. This increases number of connections and general complexity. Wouldn't it be enough to send changed connection status (or whatever would be needed) in notification? I would definitely go for it if it is possible.

If we _really_ need to let one submanager know about other, you can pass one to another during initialization (for example pass `file` submanager to `avatar`).

dvor:
Also please add some tests to following methods

> OCTManagerTests

``` objc
- (void)testInit
- (void)testSubmanagerDataSource
```

> OCTSubmanagerFriendsTests

``` objc
- (void)testFriendConnectionStatusChanged // here you can mock NSNotificationManager
```

stal888:
I'm thinking the communication would have to go both ways.

i.e OCTSubmanagerFiles receives an avatar request, passes it to OCTSubmanagerAvatars, then OCTSubmanagerAvatars has to register the OCTActiveFile with the file manager. Notifications would make this kind of use case harder to follow.

I also think that exposing every submanager up front in the same place is better than hiding references inside other submanagers.

dvor:
Well, okay, let see how it goes. Maybe will refactor it later.

objcToxTests/OCTSubmanagerFriendsTests.m
```diff
@@ -331,6 +331,11 @@ - (void)testFriendConnectionStatusChanged
     [self.submanager tox:self.tox friendConnectionStatusChanged:OCTToxConnectionStatusTCP friendNumber:kFriendNumber];
     XCTAssertEqual(friend.connectionStatus, OCTToxConnectionStatusTCP);
     XCTAssertTrue(friend.isConnected);
+
+    id mockNotificationCenter = OCMClassMock([NSNotificationCenter class]);
+    OCMStub([self.dataSource managerGetNotificationCenter]).andReturn(mockNotificationCenter);
+    [self.submanager tox:self.tox friendConnectionStatusChanged:OCTToxConnectionStatusUDP friendNumber:kFriendNumber];
+    OCMVerify([mockNotificationCenter postNotificationName:kOCTFriendConnectionStatusChangeNotificationName object:friend]);
```
dvor: It seems that `friend` is for some reason captured by OCMock. Because friend has strong reference to Realm, it's get captured too. This leads to exception in next test. I'm not sure what to do with that, need more investigation.

stal888:
The new test uses a real notification center.

dvor:
> P.S. can this be merged into file-transfers too?

Done

stal888:
Thanks

PR #122, opened 2015-08-20
Getting to compile

Chuongv:
Okay, I finally got it to compile. However I can't `OCTSubmanagerCallsTest` to pass since it fails in the setup:

`Cannot set schema version for Realms that are already open`

Is there something I am missing?

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -4,7 +4,7 @@ source 'https://github.com/CocoaPods/Specs.git'
 inhibit_all_warnings!
 
 def common_pods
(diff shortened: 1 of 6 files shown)


dvor:
I had similar issue. There are several possible reasons for that:
- realm is captured somewhere an survives between tests
- realm object is captured. It has strong reference to realm, so it got captured too
- OCTRealmManager is created several times in a test. It calls migration code two times and this crashes app.

I'll take a look at issue today.

PR #123, opened 2015-08-20
Fix tests for ocmock in octsubmanagercalls

Chuongv:
I figured it out, whenever you use OCMExpect/OCMVerify, I think it holds a reference to the realm object. Tests aren't as accurate as before but not much is lost.

### DIFF
--- a/Tests/OCTSubmanagerCallsTests.m
+++ b/Tests/OCTSubmanagerCallsTests.m
@@ -92,6 +92,10 @@
 - (void)tearDown
 {
     // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [self.dataSource stopMocking];
+    [self.mockedAudioEngine stopMocking];
+    [self.mockedVideoEngine stopMocking];
+    [self.mockedToxAV stopMocking];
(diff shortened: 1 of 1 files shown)


dvor:
Oh that's nice, good job.

dvor:
@Chuongv there is one more iOS test failing - `testIsAudioRunning` https://circleci.com/gh/Antidote-for-Tox/objcTox/68

Chuongv:
Thanks, looking into it now.

Chuongv:
Hmm it passes when I just run tests from xcode, but if I try and run the test individually, I get "The run destination iPhone 6 is not valid for tests you have chosen to perform." And if I run xctool, I get the same error as circle.

dvor:
Fixed that stuff, the problem was with initialized variable, see https://github.com/Antidote-for-Tox/objcTox/commit/920699d7c6d0fc7b3f7d3883a99c801c4fcf84eb

> "The run destination iPhone 6 is not valid for tests you have chosen to perform."

This is due to some other issue with project configuration I guess

PR #124, opened 2015-08-21
pull a base class out of OCTActiveFile

stal888:
Can't write tests while I'm breaking all the internals like this. After this I'll do file tests, then avatars, then tests for avatars, and finally put it all in the demo client.

### DIFF
--- a/Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
+++ b/Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
@@ -177,7 +177,7 @@ void OCTExceptFileNotInbound(void)
     msg.messageFile = fmsg;
 
     OCTActiveFile *send = [self createActiveFileForFriend:f message:fmsg provider:file isOutgoing:YES];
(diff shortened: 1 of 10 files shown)


Classes/Private/Manager/Submanagers/OCTSubmanagerFiles.m
```diff
@@ -468,8 +472,9 @@ - (void)     tox:(OCTTox *)tox friendConnectionStatusChanged:(OCTToxConnectionSt
         RBQFetchRequest *get = [[self.dataSource managerGetRealmManager] fetchRequestForClass:[OCTMessageAbstract class]
                                                                                 withPredicate:[NSPredicate predicateWithFormat:@"messageFile.fileState == %d && sender == nil", OCTMessageFileStateInterrupted]];
         RLMResults *objs = [get fetchObjects];
+        NSMutableArray *staticCopy = [objs valueForKey:@"self"];
```
dvor: Afaik `valueForKey:` returns `NSArray`, not the mutable one.

stal888:
I should get in the habit of building things before pushing

Issue #119, opened 2015-08-19
Video should not scale up when drawing image


Chuongv:
If user requests a UIView and uses autolayout to stretch the view, the video view should not scale to it. It should rather keep it's proportion and stay in the center.

PR #125, opened 2015-08-21
Fix video proportions keeping aspect fit

Chuongv:
https://github.com/Antidote-for-Tox/Antidote/issues/122
https://github.com/Antidote-for-Tox/objcTox/issues/119

### DIFF
--- a/Classes/Private/Manager/Video/OCTVideoView.m
+++ b/Classes/Private/Manager/Video/OCTVideoView.m
@@ -8,6 +8,7 @@
 
 #import "OCTVideoView.h"
 @import Foundation;
+@import AVFoundation;
 
 @interface OCTVideoView ()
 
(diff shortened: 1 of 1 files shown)


Classes/Private/Manager/Video/OCTVideoView.m
```diff
@@ -47,8 +48,16 @@ - (void)setImage:(CIImage *)image
 - (void)drawRect:(CGRect)rect
 {
     if (self.image) {
-        CGFloat scale = self.window.screen.scale;
-        CGRect destRect = CGRectApplyAffineTransform(self.bounds, CGAffineTransformMakeScale(scale, scale));
+
+        glClearColor(0, 0.0, 0.0, 1.0);
+        glClear(GL_COLOR_BUFFER_BIT);
```
dvor, line 53: Should we do that every 'drawRect:' method call or maybe we can call it only once?

dvor:
Also please rebase on top of audio to fix circleci

Classes/Private/Manager/Video/OCTVideoView.m
Chuongv, line 53: Hmm yea I could probably optimize this a bit. Basically check if the latest `self.image.extent` has changed, and if it did, call `glClear`

Chuongv:
It is already rebased on top of audio. Did you want me to rebase on top of `getting-latest-changes-to-audio` branch now?

dvor:
Oh, sorry, I missed that there is old `toxcore-ios` pod on audio branch. And circle is unhappy with it.

It is strange that `audio` is building at all (perhaps it uses cached pods).

Classes/Private/Manager/Video/OCTVideoView.m
Chuongv, line 53: Hmm doing the above doesn't work since there's still some artifacts left over after clearing it. It probably has to be cleared more than once

dvor, line 53: By "above" you mean this?

> Hmm yea I could probably optimize this a bit. Basically check if the latest self.image.extent has changed, and if it did, call glClear

Code existing in pull request works?

Chuongv, line 53: Yea the pull request as is works. I actually think glClear before every draw is not bad. According to this I think https://developer.apple.com/library/ios/qa/qa1650/_index.html

dvor, line 53: Okay then

Issue #85, opened 2015-08-03
Allow user to pick either front or rear camera


Chuongv:
Allow user to pick either front or rear camera during video calls. This should be implemented at `OCTSubmanagerCalls` and `OCTVideoEngine`.

Right now the default is the front facing camera.

PR #129, opened 2015-08-27
Add ability to switch front and back camera position

Chuongv:
https://github.com/Antidote-for-Tox/objcTox/issues/85

### DIFF
--- a/Classes/Private/Manager/Video/OCTVideoEngine.m
+++ b/Classes/Private/Manager/Video/OCTVideoEngine.m
@@ -81,7 +81,9 @@ static const OSType kPixelFormat = kCVPixelFormatType_420YpCbCr8BiPlanarVideoRan
         return NO;
     }
 
-    [self.captureSession addInput:videoInput];
+    if ([self.captureSession canAddInput:videoInput]) {
+        [self.captureSession addInput:videoInput];
+    }
 
     self.dataOutput.alwaysDiscardsLateVideoFrames = YES;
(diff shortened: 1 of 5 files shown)


Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -100,6 +100,46 @@ - (BOOL)setupWithError:(NSError **)error
...
+
+    AVCaptureDevice *camera = [self getDeviceForPosition:position];
+
+    AVCaptureDeviceInput *videoInput = [AVCaptureDeviceInput deviceInputWithDevice:camera error:error];
+
+    if (! videoInput) {
+        return NO;
+    }
+
+    if (! [self.captureSession canAddInput:videoInput]) {
```
dvor, line 133: I think you should check this in `setupWithError:` method too.

> ``` objc
> - (void)addInput:(AVCaptureInput *)input
> ```
> 
> You can only add an input to a session using this method if canAddInput: returns YES.

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -100,6 +100,46 @@ - (BOOL)setupWithError:(NSError **)error
...
+        // fill error?
```
dvor: I think it is okay to ignore it as there should be no error in all normal cases.

dvor:
Circleci seems to be broken. Will fix this later.

Issue #140, opened 2015-10-02
Call tox_iteration_interval at every tox_iterate


Chuongv:
Similar issue to https://github.com/Antidote-for-Tox/objcTox/issues/113 instead with OCTTox.

Currently when OCTTox starts, we grab the tox_iteration_interval once and set the timer with that number. This is incorrect, the iteration interval should be updated at every tox_iterate

PR #141, opened 2015-10-02
Call tox_iterate_interval after every tox_iterate

Chuongv:
https://github.com/Antidote-for-Tox/objcTox/issues/140


dvor:
Cool, thanks. I've missed that for some reason.

Issue #88, opened 2015-08-04
Automatically adjust video bitrate/resolution when network quality degrades


Chuongv:
When network quality goes bad, objcTox should be either lower the resolution or ask toxav to lower the video bitrate.

Chuongv:
With the new toxAV, we receive a call back that actually suggests the bitrate we should move to. However the call back will never suggest a higher bit rate. So we will need to set a timer that will set the bit rate higher after it has been lowered. If we move it to high, toxAV will suggest a more appropriate bit rate.

So for example, we start a call with 48kb/s bit rate audio. 
ToxAV callback with 24kb/s for audio. 
ObjcTox sets bitrate to 24kb/s for audio.
ObjcTox sets timer to attempt to move bitrate back to 48 after 3 minutes.
3 minutes pass.. ObjcTox sets bitrate to 48.
ToxAV callback with 36 kb/s.
ObjcTox sets bitrate to 36 kb/s

DanTheBritish:
3 minutes is a long time, could it not be 1 minute (or even 30 seconds)?

dvor:
@Chuongv what will happen when we'll set bit rate to too high value? Would toxav immediately lower it? In this case we can try to increase it really often (like 10 seconds maybe?).

Chuongv:
According to mannol, if we set it too high, it will suggest a new bit rate after 2 seconds. So yea we could make it increase often every 10 seconds.

Chuongv:
Depends on https://github.com/irungentoo/toxcore/issues/1469

PR #143, opened 2015-10-15
Update to latest toxAV from mannol

Chuongv:
Some new enums were added from toxav.

Did `$ xctool -workspace objcTox.xcworkspace -scheme iOSDemo -sdk iphonesimulator CODE_SIGNING_REQUIRED=NO GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES GCC_GENERATE_TEST_COVERAGE_FILES=YES clean test`

But got one error:
`Unable to read build settings for target 'iOSDemoTests'.  It's likely that the scheme references a non-existent target.`

`xcodebuild: error: The test action requires that the name of a scheme in the project is provided using the "-scheme" option. The "-list" option can be used to find the names of the schemes in the project.`

However all tests pass when running them in xcode

### DIFF
--- a/Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
+++ b/Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
@@ -480,9 +480,9 @@ const OCTToxAVVideoBitRate kDefaultVideoBitRate = 2000;
     [self.audioEngine provideAudioFrames:pcm sampleCount:sampleCount channels:channels sampleRate:sampleRate fromFriend:friendNumber];
 }
 
-- (void)toxAV:(OCTToxAV *)toxAV bitrateStatusForFriendNumber:(OCTToxFriendNumber)friendNumber
- audioBitRate:(OCTToxAVAudioBitRate)audioBitrate
- videoBitRate:(OCTToxAVVideoBitRate)videoBitrate
+- (void)   toxAV:(OCTToxAV *)toxAV bitrateStatusForFriendNumber:(OCTToxFriendNumber)friendNumber
+    audioBitRate:(OCTToxAVAudioBitRate)audioBitrate
+    videoBitRate:(OCTToxAVVideoBitRate)videoBitrate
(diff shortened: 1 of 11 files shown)


Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
```diff
@@ -514,6 +480,11 @@ - (void)   toxAV:(OCTToxAV *)toxAV
     [self.audioEngine provideAudioFrames:pcm sampleCount:sampleCount channels:channels sampleRate:sampleRate fromFriend:friendNumber];
 }
 
+- (void)   toxAV:(OCTToxAV *)toxAV bitrateStatusForFriendNumber:(OCTToxFriendNumber)friendNumber
+    audioBitRate:(OCTToxAVAudioBitRate)audioBitrate
+    videoBitRate:(OCTToxAVVideoBitRate)videoBitrate
+{}
```
dvor: Why objcTox isn't handling this callback?

dvor:
>  Unable to read build settings for target 'iOSDemoTests'. It's likely that the scheme references a non-existent target.

I'll take a look. Created an issue https://github.com/Antidote-for-Tox/objcTox/issues/144

Classes/Private/Manager/Submanagers/OCTSubmanagerCalls.m
Chuongv: Yea you are right, this should be done at objcTox level. I'll move this.

dvor: Hm... I think that it should be done **not** in OCTToxAV, but on more higher level. OCTSubmanagerCalls would be perfect place for that. So everything was fine, no need to remove the callback. I was just writing about those empty parentheses.

Can you please get it back (and add a comment to the #88 here)? :-)
Sorry for confusing.

dvor:
Cool, thanks!

PR #149, opened 2015-12-23
Fix negative stride handling and add tests for it

stal888:
As iphy pointed out in IRC this morning, nobody was doing it correctly.

Test image generated using uTox code: https://gist.github.com/stal888/7d267c189c8199f51e12


dvor:
Great, thanks!

Tests/OCTVideoEngineTests.m
```diff
@@ -145,4 +151,81 @@ - (void)testReceiveVideoFrame
     });
 }
 
+- (void)_performYUVCheckWithPlanes:(uint8_t * [3])planes strides:(OCTToxAVStrideData[3])strides
```
dvor: Can you remove `_` from method name? So `_performYUVCheckWithPlanes` -> `performYUVCheckWithPlanes` :-)

Tests/OCTVideoEngineTests.m
```diff
@@ -145,4 +151,81 @@ - (void)testReceiveVideoFrame
...
+        XCTAssertTrue(pass);
+    });
+}
+
+- (void)testReceivePackedVideoFrame
+{
+    uint8_t *planes[3] = {test_str83p_y, test_str83p_u, test_str83p_v};
+    OCTToxAVStrideData strides[3] = {30, 15, 15};
+
+    [self _performYUVCheckWithPlanes:planes strides:strides];
```
dvor: Tests are failing on OS X as appropriate functionality isn't implemented yet. For now you can wrap this four tests in following macro:

``` objc
- (void)testReceivePackedVideoFrame
{
#if TARGET_OS_IPHONE
    uint8_t *planes[3] = {test_str83p_y, test_str83p_u, test_str83p_v};
    OCTToxAVStrideData strides[3] = {30, 15, 15};

    [self _performYUVCheckWithPlanes:planes strides:strides];
#else
#warning TODO audio OSX
#endif
}
```

stal888: The CIImage code works just fine on OS X so I've removed the guard around that instead of putting new ones on the tests.

dvor:
:+1:

Issue #118, opened 2015-08-19
Add configuration and sample project for OS X 


dvor:

dvor:
@Chuongv will assign it to you if you like

dvor:
Already done

PR #150, opened 2015-12-23
Chuongv/chats

Chuongv:
Part of https://github.com/Antidote-for-Tox/objcTox/issues/118

This completes the messaging portion of the demo. Test it out and let me know if you see any bugs.

### DIFF
--- a/OSXDemo/OCTConversationViewController.m
+++ b/OSXDemo/OCTConversationViewController.m
@@ -254,6 +254,8 @@ static NSString *const kCellIdent = @"cellIdent";
 
     if (chat) {
         NSIndexSet *indexSet = [[NSIndexSet alloc] initWithIndex:0];
+        // workaround for deadlock in objcTox
+        // https://github.com/Antidote-for-Tox/objcTox/issues/51
         dispatch_async(dispatch_get_main_queue(), ^{
             [self.chatsTableView selectRowIndexes:indexSet byExtendingSelection:NO];
         });
--- a/OSXDemo/OCTConversationViewController.m
+++ b/OSXDemo/OCTConversationViewController.m
@@ -77,28 +77,6 @@ static NSString *const kCellIdent = @"cellIdent";
     [self.manager.chats removeChatWithAllMessages:chat];
 }
 
-- (void)tableViewSelectionDidChange:(NSNotification *)notification
-{
-    if (notification.object != self.chatsTableView) {
-        return;
-    }
-
-    NSInteger selectedRow = self.chatsTableView.selectedRow;
-
-    if (selectedRow < 0) {
-        return;
-    }
-
-    NSIndexPath *path = [NSIndexPath indexPathForRow:selectedRow inSection:0];
-    OCTChat *chat = [self.chatResultsController objectAtIndexPath:path];
-
-    [self updateConversationControllerForChat:chat];
-
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self.conversationTableView reloadData];
-    });
-}
-
 #pragma mark - NSTextFieldDelegate
 
 - (IBAction)chatTextFieldEntered:(NSTextField *)sender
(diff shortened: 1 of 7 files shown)


OSXDemo/OCTConversationViewController.m
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    NSIndexPath *path = [NSIndexPath indexPathForRow:selectedRow inSection:0];
+    OCTChat *chat = [self.chatResultsController objectAtIndexPath:path];
+
+    [self updateConversationControllerForChat:chat];
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.conversationTableView reloadData];
+    });
```
dvor: What is reason for using `dispatch_async` here?

OSXDemo/OCTConversationViewController.m
```diff
@@ -0,0 +1,264 @@
...
+
+- (void) controller:(RBQFetchedResultsController *)controller
+    didChangeObject:(RBQSafeRealmObject *)anObject
+        atIndexPath:(NSIndexPath *)indexPath
+      forChangeType:(RBQFetchedResultsChangeType)type
+       newIndexPath:(NSIndexPath *)newIndexPath
+{
+    NSTableView *tableView = (self.chatResultsController == controller) ? self.chatsTableView : self.conversationTableView;
+
+    dispatch_async(dispatch_get_main_queue(), ^{
```
dvor: What's the reason for using `dispatch_async` here?

dvor:
Looks good!

OSXDemo/OCTConversationViewController.m
Chuongv: If I don't, it runs off the RBQFetchResults queue, which causes some UI issues since we are touching the NSTableView.

OSXDemo/OCTConversationViewController.m
Chuongv: To ensure any UI stuff goes on the main thread. Though I guess it's not necessary needed here since this method is always under the main queue?

dvor: Yep, table view calls it's delegate methods on main thread

OSXDemo/OCTConversationViewController.m
dvor: RBQ already calls this method on main thread. The only issue we had was dealloc, but it is not an issue in this case.

dvor: s/dealloc/deadlock
_need more sleep_

Chuongv: > RBQ already calls this method on main thread.  

I don't think that's always true though. I see it called on the OCTRealmManager queue. I placed a breakpoint and see it called off there. It's giving me a deadlock if I don't dispatch to main queue in that case. Either way, I was able to fix this by instead just dispatching only `[self.chatsTableView selectRowIndexes:indexSet byExtendingSelection:NO];` to the main thread under `- (void)selectFirstChat`

dvor: > I see it called on the OCTRealmManager queue.

Are you sure about that?

It should call it only on main queue ([code](https://github.com/Antidote-for-Tox/objcTox/blob/master/Classes/Public/Manager/RBQFetchedResultsController/RBQFetchedResultsController.m#L1527-L1533)). Also I've checked, it works for me without dispatch_async

Chuongv: Maybe I am reading it wrong?

screenshot

dvor: Well, "Thread 1" means main queue. :-)
You can click on stack trace to see what's actually happening there. 

screen shot 2015-12-24 at 01 02 05

OSXDemo/OCTConversationViewController.m
```diff
@@ -0,0 +1,262 @@
...
+}
+
+- (void)selectFirstChat
+{
+    NSIndexPath *path = [NSIndexPath indexPathForRow:0 inSection:0];
+    OCTChat *chat = [self.chatResultsController objectAtIndexPath:path];
+
+    if (chat) {
+        NSIndexSet *indexSet = [[NSIndexSet alloc] initWithIndex:0];
+        dispatch_async(dispatch_get_main_queue(), ^{
```
dvor, line 259: So I think there is no need in this `dispatch_async`

OSXDemo/OCTConversationViewController.m
Chuongv: Ahh I see, I guess learn something new everyday :)

OSXDemo/OCTConversationViewController.m
Chuongv, line 259: hmm if remove this and I go delete a chat when there is more than 1 chat available, the program locks up.

dvor, line 259: I see. Can you then add famous comment (hopefully this will be fixed one day).

``` objc
// workaround for deadlock in objcTox
// https://github.com/Antidote-for-Tox/objcTox/issues/51
```

dvor:
Thanks!

PR #153, opened 2015-12-29
Make all A/V stuff compatible with OS X.

stal888:
feat. Mac demo by @Chuongv 

Adds Mac specific bits to OCTVideoView, and a rewritten AudioEngine with new device switching API. All this should still work on iOS, but I haven't tried it.


Chuongv:
> All this should still work on iOS, but I haven't tried it.

I haven't tried it yet either so let's hold off on the merge until one of us does it. I'll take a look soon hopefully.

However, the mac audio/video works great :+1:

Chuongv:
I am not hearing sound when using the iOS demo to call the objcToxBot.

Saw this in the logs 
`2015-12-29 15:30:03:211 iOSDemo[7399:b07] Error starting audio flow Error Domain=NSOSStatusErrorDomain Code=-66680 "Consult the CoreAudio header files/google for the meaning of the error code." UserInfo={NSLocalizedDescription=Consult the CoreAudio header files/google for the meaning of the error code.}`

Looking up 66680 it's:
`The specified audio hardware device could not be located.`

Also video takes a longer than usual to start up, 10-15 seconds?

stal888:
Might have fixed it... hard to tell because startAudioFlow wraps the entire creation process of the OCTAudioQueue.

> Also video takes a longer than usual to start up, 10-15 seconds?

I think this is caused by toxav.

dvor:
Some tests are failing. Also uncrustify is unhappy, please run `./run-uncrustify.sh --apply` to make build green.

Chuongv:
@stal888 

Tested a call between objcTox mac to objcTox iOS:
The iphone is able to hear what the mac is sending. So it seems that iOS is able to retrieve the audio frames and play them correctly, just not able to send audio though.

Video works on receiving for mac, but only if the iphone is on landscape mode. It may be a autolayout issue with my mac demo though.

Chuongv:
Audio and video works now on iOS. :+1:

stal888:
I'm going to fix the mocking in OCTAudioEngineTests soon.

stal888:
@dvor Looks like uncrustify can't handle ternary operators. I guess it'll have to stay broken for now.

dvor:
You can apply it as it is. I think it is fine to have following in 5 places, as it gives same style all over the code.

``` objc
NSData *data = (rdlen > 0) ? ([NSData dataWithBytes : rdata length:rdlen]) : nil;
```

stal888:
Ok. Previous commit amended.

stal888:
> Failure: A command timed out during your tests

Should be fine...

dvor:
Sorry for long delay, will review it now

Classes/Private/Manager/Audio/OCTAudioEngine.h
```diff
@@ -35,12 +38,14 @@
 - (BOOL)stopAudioFlow:(NSError **)error;
 
 /**
- * Send the audio to the speaker
- * @param speaker YES to send audio to speaker, NO to reset to default.
- * @param error Pointer to error object.
- * @return YES if successful override, otherwise NO.
+ * Set the input device (not available on Mac OS X).
+ * @param inputDeviceID Core Audio's unique ID for the device. See
```
dvor: "See" what?

Classes/Private/Manager/Audio/OCTAudioEngine.h
```diff
@@ -35,12 +38,14 @@
 - (BOOL)stopAudioFlow:(NSError **)error;
 
 /**
- * Send the audio to the speaker
- * @param speaker YES to send audio to speaker, NO to reset to default.
- * @param error Pointer to error object.
- * @return YES if successful override, otherwise NO.
+ * Set the input device (not available on Mac OS X).
```
dvor: Commen here is wrong. Set input device is not available for iOS, not OS X

Classes/Private/Manager/Audio/OCTAudioEngine.h
```diff
@@ -35,12 +38,14 @@
...
+ * @param inputDeviceID Core Audio's unique ID for the device. See
+ * @param error If this method returns NO, contains more information on the
+ *              underlying error.
+ * @return YES on success, otherwise NO.
  */
-- (BOOL)routeAudioToSpeaker:(BOOL)speaker error:(NSError **)error;
+- (BOOL)setInputDeviceID:(NSString *)inputDeviceID error:(NSError **)error;
+- (BOOL)setOutputDeviceID:(NSString *)outputDeviceID error:(NSError **)error;
```
dvor: Can you also add a documentation for setOutputDeviceID?

Classes/Private/Manager/Audio/OCTAudioEngine.m
```diff
@@ -83,459 +38,162 @@ - (instancetype)init
...
+        if (aoi.enableMicrophone) {
+            [aoi.toxav sendAudioFrame:data
+                          sampleCount:samples
+                             channels:channelCount
+                           sampleRate:rate
+                             toFriend:aoi.friendNumber
+                                error:nil];
+        }
+    };
+    [self.outputQueue updateSampleRate:(Float64)self.outputSampleRate numberOfChannels:(UInt32)self.outputNumberOfChannels error:nil];
```
dvor: outputQueue can take OCTToxAVSampleRate and OCTToxAVChannels as parameters and do a cast itself. That way we'll have OCT types in more places

Classes/Private/Manager/Audio/OCTAudioEngine.m
```diff
@@ -83,459 +38,162 @@ - (instancetype)init
...
-    TPCircularBufferProduceBytes(&_outputBuffer, pcm, len);
-
-    if ((self.outputSampleRate != sampleRate) || (self.outputNumberOfChannels != channels)) {
-        NSError *error;
-        if (! [self updateOutputSampleRate:sampleRate channels:channels error:&error]) {
-            DDLogWarn(@"%@, error updateOutputSampleRate:%@", self, error);
-        }
+    else {
+        return NO;
     }
```
dvor, line 90: No need for else here, you can simply `return NO;`. (it is a matter of style, but I prefer early returns without "else" part).

Classes/Private/Manager/Audio/OCTAudioEngine.m
```diff
@@ -83,459 +38,162 @@ - (instancetype)init
...
-                         sampleCount:kSampleCount
-                            channels:kNumberOfInputChannels
-                          sampleRate:engine.inputSampleRate
-                            toFriend:engine.friendNumber
-                               error:&error];
-        TPCircularBufferConsume(&engine->_inputBuffer, minimalBytesToConsume);
-        tail = TPCircularBufferTail(&engine->_inputBuffer, &availableBytesToConsume);
+    else {
+        return YES;
     }
```
dvor: No need for else here, you can simply `return NO;`. (it is a matter of style, but I prefer early returns without "else" part).

Classes/Private/Manager/Audio/OCTAudioEngine.m
```diff
@@ -83,459 +38,162 @@ - (instancetype)init
...
+- (void)provideAudioFrames:(OCTToxAVPCMData *)pcm sampleCount:(OCTToxAVSampleCount)sampleCount channels:(OCTToxAVChannels)channels sampleRate:(OCTToxAVSampleRate)sampleRate fromFriend:(OCTToxFriendNumber)friendNumber
 {
-    OSStatus status = _AUGraphInitialize(self.processingGraph);
-    if (status != noErr) {
-        [self fillError:error
-               withCode:status
-            description:@"Initialize Graph"
-          failureReason:@"Failed to initialize Graph"];
-        return NO;
+    int32_t len = (int32_t)(channels * sampleCount * sizeof(int16_t));
```
dvor: `sizeof(OCTToxAVPCMData)` would look better here

Classes/Private/Manager/Audio/OCTAudioEngine.h
dvor: Even better - maybe we should make whole method `setInputDeviceID:error:` OS X only? That way there would be no need for this comment, for exception for iOS

dvor: Same for inputDeviceID property. It should be OS X only.

dvor: Looking at the code even more, I would create separate API for OS X and iOS. IMO no need for similar interface here. Those methods don't have any common code. Old way it would be less confusing:

``` objc
#if TARGET_OS_IPHONE

- (BOOL)routeAudioToSpeaker:(BOOL)speaker error:(NSError **)error;

#else

- (BOOL)setInputDeviceID:(NSString *)inputDeviceID error:(NSError **)error;
- (BOOL)setOutputDeviceID:(NSString *)outputDeviceID error:(NSError **)error;

#endif
```

Classes/Private/Manager/Audio/OCTAudioQueue.h
```diff
@@ -0,0 +1,75 @@
...
+                                               const void *inQualifierData,
+                                               UInt32 *ioDataSize,
+                                               void *outData);
+#endif
+
+/* no idea what to name this thing, so here it is */
+@interface OCTAudioQueue : NSObject
+
+@property (strong, nonatomic, readonly) NSString *deviceID;
+@property (strong, nonatomic) void (^sendDataBlock)(void *, OCTToxAVSampleCount, OCTToxAVSampleRate, OCTToxAVChannels);
```
dvor: Block property should be copy

Classes/Private/Manager/Audio/OCTAudioQueue.m
```diff
@@ -0,0 +1,418 @@
...
+
+static NSError *OCTErrorFromCoreAudioCode(OSStatus resultCode)
+{
+    return [NSError errorWithDomain:NSOSStatusErrorDomain
+                               code:resultCode
+                           userInfo:@{NSLocalizedDescriptionKey : @"Consult the CoreAudio header files/google for the meaning of the error code."}];
+}
+
+#if ! TARGET_OS_IPHONE
+static NSString *OCTGetSystemAudioDevice(AudioObjectPropertySelector sel, NSError **err)
```
dvor, line 82: What's the reason of using static C function instead of objc method? Here and above for `OCTErrorFromCoreAudioCode`

Classes/Private/Manager/Audio/OCTAudioQueue.m
```diff
@@ -0,0 +1,418 @@
...
+    _streamFmt.mBytesPerFrame = kBytesPerSample * kNumberOfInputChannels;
+    _streamFmt.mBitsPerChannel = kBitsPerByte * kBytesPerSample;
+    _streamFmt.mFramesPerPacket = kFramesPerPacket;
+    _streamFmt.mBytesPerPacket = kBytesPerSample * kNumberOfInputChannels * kFramesPerPacket;
+    _isOutput = NO;
+    _deviceID = devID;
+
+    TPCircularBufferInit(&_buffer, kBufferLength);
+    OSStatus res = 0;
+    if ((res = [self createAudioQueue]) != 0) {
```
dvor:
``` objc
OSStatus res = [self createAudioQueue];
if (res != 0) {
```

would be more readable

Classes/Private/Manager/Audio/OCTAudioQueue.m
```diff
@@ -0,0 +1,418 @@
...
+
+@implementation OCTAudioQueue {
+    AudioQueueBufferRef _AQBuffers[kNumberOfAudioQueueBuffers];
+}
+
+- (instancetype)initWithInputDeviceID:(NSString *)devID error:(NSError **)error
+{
+#if TARGET_OS_IPHONE
+    AVAudioSession *session = [AVAudioSession sharedInstance];
+    _streamFmt.mSampleRate = session.sampleRate;
```
dvor, line 137: Why we are using sampleRate from AVAudioSession here and not the default one?

Classes/Private/Manager/Audio/OCTAudioEngine+Private.h
```diff
@@ -11,68 +11,21 @@
...
-                             AudioBufferList *ioData);
-
-OSStatus outputRenderCallBack(void *inRefCon,
-                              AudioUnitRenderActionFlags *ioActionFlags,
-                              const AudioTimeStamp *inTimeStamp,
-                              UInt32 inBusNumber,
-                              UInt32 inNumberFrames,
-                              AudioBufferList *ioData);
+extern int kBufferLength;
+extern int kNumberOfInputChannels;
```
dvor: This number is used for output as well. `kNumberOfInputChannels` -> `kNumberOfChannels`.

Classes/Private/Manager/Audio/OCTAudioQueue.m
```diff
@@ -0,0 +1,418 @@
...
+    TPCircularBufferInit(&_buffer, kBufferLength);
+    OSStatus res = 0;
+    if ((res = [self createAudioQueue]) != 0) {
+        if (error) {
+            *error = OCTErrorFromCoreAudioCode(res);
+        }
+        return nil;
+    }
+
+    return self;
```
dvor: You can avoid code duplication by moving shared code to private init method. Then you can call it from both inits. Something like that

``` objc
- (instancetype)initWithDeviceID:(NSString *)devID
                        isOutput:(BOOL)isOutput
                      sampleRate:(Float64)sampleRate
                     formatFlags:(AudioFormatFlags)formatFlags
                           error:(NSError **)error
```

Classes/Private/Manager/Audio/OCTAudioQueue.h
```diff
@@ -0,0 +1,75 @@
...
+@property (assign, nonatomic, readonly) BOOL running;
+
+- (instancetype)initWithInputDeviceID:(NSString *)devID error:(NSError **)error;
+- (instancetype)initWithOutputDeviceID:(NSString *)devID error:(NSError **)error;
+
+- (TPCircularBuffer *)getBufferPointer;
+- (BOOL)updateSampleRate:(Float64)sampleRate numberOfChannels:(UInt32)numberOfChannels error:(NSError **)err;
+- (BOOL)setDeviceID:(NSString *)deviceID error:(NSError **)err;
```
dvor, line 72: This method can be wrapper in OS X only macro. No need to confuse iOS user

Classes/Private/Manager/Video/OCTVideoEngine.h
```diff
@@ -30,12 +30,11 @@
...
  * Use a different camera for input.
- * @param front YES to use the front camera, NO to use the
- * rear camera. Front camera is used by default.
- * @error Pointer to error object.
+ * @param camera The camera's AVFoundation device ID.
+ * @param error Pointer to error object.
  * @return YES on success, otherwise NO.
  */
-- (BOOL)switchToCameraFront:(BOOL)front error:(NSError **)error;
+- (BOOL)switchToCamera:(NSString *)camera error:(NSError **)error;
```
dvor: Here same as for OCTAudioEngine. I think it would be better to have separate methods for OS X and iOS (as for OS X we can pas any camera string, for iOS we can only switch between front and rear cameras).

Classes/Private/Manager/Video/OCTVideoView.m
```diff
@@ -18,41 +18,51 @@ @interface OCTVideoView ()
...
-#else
-#warning TODO audio OSX
-#endif
     });
 
-#if TARGET_OS_IPHONE
     self.enableSetNeedsDisplay = NO;
-#else
-#warning TODO audio OSX
+    // #warning TODO audio OSX
```
dvor: Do we need to do something more? Otherwise you can remove comment.

Classes/Private/Manager/Audio/OCTAudioQueue.m
stal888, line 137: I copied this behaviour from the old audio engine. It seems to be recommended by Apple: https://developer.apple.com/library/ios/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/OptimizingForDeviceHardware/OptimizingForDeviceHardware.html#//apple_ref/doc/uid/TP40007875-CH6-SW2

Classes/Private/Manager/Audio/OCTAudioQueue.m
stal888, line 82: We don't need self

Classes/Private/Manager/Audio/OCTAudioQueue.m
Chuongv: Instead of comparing it to `0` I think it would be better to compare it to `noErr`. Just to keep it consistent with the rest of the OCTAudioEngine.

```
OSStatus res = [self createAudioQueue];
if (res != noErr) {
```

Chuongv: Err nevermind, all the `noErr` have been remove.

Classes/Private/Manager/Audio/OCTAudioQueue.m
dvor, line 82: What's the issue with using self? Both `OCTErrorFromCoreAudioCode` and `OCTGetSystemAudioDevice` are used only within class.

Classes/Private/Manager/Audio/OCTAudioQueue.m
dvor, line 137: Can you please add a comment about that in the code (with a link)?

Classes/Private/Manager/Audio/OCTAudioQueue.m
stal888, line 82: No, I mean those functions don't need the self parameter. They're not conceptually related to the class so I moved them out.

Classes/Private/Manager/Audio/OCTAudioEngine.h
```diff
@@ -11,6 +11,9 @@
 
 @interface OCTAudioEngine : NSObject
 
+@property (strong, nonatomic, readonly) NSString *inputDeviceID;
+@property (strong, nonatomic, readonly) NSString *outputDeviceID;
```
dvor, line 15: These two properties should also be a part of MacDevice

dvor, line 15: Well, I can do that myself.

dvor:
Thanks, great job!

Issue #151, opened 2015-12-23
OCTFriend.lastSeenOnline is always nil


dvor:
It seems that OCTFriend.lastSeenOnline method always returns nil

dvor:
Fixed by #155

PR #155, opened 2016-01-02
Last Seen online bug fix

Chuongv:
For https://github.com/Antidote-for-Tox/objcTox/issues/151

The lastSeenOnlineInterval is always updated after a friend's connection
status goes to offline. If friend goes online, the last seen online will
be nil. Subsequently the lastSeenInterval is set to 0.

### DIFF
--- a/Classes/Private/Manager/Submanagers/OCTSubmanagerFriends.m
+++ b/Classes/Private/Manager/Submanagers/OCTSubmanagerFriends.m
@@ -106,6 +106,8 @@ NSString *const kOCTFriendConnectionStatusChangeNotificationName = @"kOCTFriendC
             friend.isConnected = NO;
             friend.connectionStatus = OCTToxConnectionStatusNone;
             friend.isTyping = NO;
+            NSDate *dateOffline = [[self.dataSource managerGetTox] friendGetLastOnlineWithFriendNumber:friend.friendNumber error:nil];
+            friend.lastSeenOnlineInterval = [dateOffline timeIntervalSince1970];
         }
     }];
 
@@ -216,9 +218,6 @@ NSString *const kOCTFriendConnectionStatusChangeNotificationName = @"kOCTFriendC
             NSDate *dateOffline = [tox friendGetLastOnlineWithFriendNumber:friendNumber error:nil];
             NSTimeInterval timeSince = [dateOffline timeIntervalSince1970];
             theFriend.lastSeenOnlineInterval = timeSince;
-        }
-        else {
-            theFriend.lastSeenOnlineInterval = 0;
         }
     }];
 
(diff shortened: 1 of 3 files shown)


Classes/Private/Manager/Submanagers/OCTSubmanagerFriends.m
```diff
@@ -211,6 +211,15 @@ - (void)tox:(OCTTox *)tox friendConnectionStatusChanged:(OCTToxConnectionStatus)
...
         theFriend.isConnected = (status != OCTToxConnectionStatusNone);
         theFriend.connectionStatus = status;
+
+        if (! theFriend.isConnected) {
+            NSDate *dateOffline = [tox friendGetLastOnlineWithFriendNumber:friendNumber error:nil];
+            NSTimeInterval timeSince = [dateOffline timeIntervalSince1970];
+            theFriend.lastSeenOnlineInterval = timeSince;
+        }
+        else {
+            theFriend.lastSeenOnlineInterval = 0;
```
dvor: Hm, I don't get it. Why do you want to set `lastSeenOnlineInterval` to 0? It should have some value if friend was online at least once.

I see following cases:
1. If friend is online, we don't care about `lastSeenOnlineInterval`. Our documentation says that it `Contains actual information in case if friend has connectionStatus offline.`
2. Friend was online, now he goes offline. We should update `lastSeenOnlineInterval` with new value.
3. Friend was online, the app was killed. There was no chance of updating `lastSeenOnlineInterval` value for us. In that case we can update interval on next launch.

So, here you can leave only:

``` objc
if (! theFriend.isConnected) {
    NSDate *dateOffline = [tox friendGetLastOnlineWithFriendNumber:friendNumber error:nil];
    theFriend.lastSeenOnlineInterval = [dateOffline timeIntervalSince1970];
}
```

And in `configure` method it would be also nice to update interval:

``` objc
- (void)configure
{
    ...

     // update some of friend properties
    [realmManager updateObjectsWithoutNotification:^{
        for (OCTFriend *friend in friendsArray) {
            friend.status = OCTToxUserStatusNone;
            friend.isConnected = NO;
            friend.connectionStatus = OCTToxConnectionStatusNone;
            friend.isTyping = NO;
            NSDate *dateOffline = [tox friendGetLastOnlineWithFriendNumber:friendNumber error:nil];
            friend.lastSeenOnlineInterval = [dateOffline timeIntervalSince1970];
        }
    }];

    ...
}
```

Chuongv: Ah thanks. I keep forgetting that our tox file has the latest last seen date.

dvor:
Cool

PR #156, opened 2016-01-05
get CVPixelBuffer's stride instead of assuming

stal888:
It was causing scrambled video.


stal888:
I'll fix the tests soon™

dvor:
Thanks

Issue #148, opened 2015-12-15
Bug: Delete friend not found, OSX


Chuongv:
In `OCTSubmanagerFriends`, if a user  tries to delete a friend and tox returns "Friend not found", objcTox should remove that friend from realm.

I am not sure how I was able to delete a tox friend without deleting it from realm in the first place. But it's present somehow.

dvor:
More interesting is to find the reason of having friend in realm only. Maybe tox wasn't saved after adding friend? Or a crash occured somewhere?

Chuongv:
Ah I figured it out. I did an Xcode clean for mac osx, which does a clean of the tox data, but it doesn't get rid of the realm database. So I had a new tox profile but with a realm database for a different tox file loaded.

dvor:
What do you mean by "Xcode clean"?

.tox file should be stored together with .realm database in one place.

Chuongv:
Ah yea you are right. They are saved together. I'll try and reproduce it again.

Chuongv:
Ah finally figured it out :smile: . If you run the mac tests, it will clear out the tox file in the documents folder but leave the database behind. So when you load up the mac app again from objcTox, it will create a new tox profile but then load the old database.

Chuongv:
Closing this due to PR #157

PR #157, opened 2016-01-17
Move saved database to test location

Chuongv:
Related to: https://github.com/Antidote-for-Tox/objcTox/issues/148
Before this, the test location was creating a new tox file in the
same default directory as the default configuration for OCTFileStorage. Thus each time I ran tests, I would get a new tox profile but with my old realm database.


dvor:
Thanks!

Issue #154, opened 2016-01-01
Kill toxav before killing tox


stal888:
~~Sometimes~~ they are deallocated in the wrong order, causing a segfault when setting OCTManager to nil.

Reproduction steps:
- Be in a call
- Make OCTManager dealloc

Relevant log: 

```
2016-01-01 13:53:17.634 Nao[58565:32508349] <OCTTox: 0x6000000400f0>: stop method called
2016-01-01 13:53:17.635 Nao[58565:32508349] <OCTTox: 0x6000000400f0>: stopped
2016-01-01 13:53:17.635 Nao[58565:32508349] <OCTTox: 0x6000000400f0>: stop method called
2016-01-01 13:53:17.635 Nao[58565:32508349] <OCTTox: 0x6000000400f0>: tox isn't running, nothing to stop
2016-01-01 13:53:17.635 Nao[58565:32508349] <OCTTox: 0x6000000400f0>: dealloc called, tox killed
2016-01-01 13:53:17.635 Nao[58565:32508349] OCTAudioQueue stop
2016-01-01 13:53:17.643 Nao[58565:32508349] Freed buffers
2016-01-01 13:53:17.644 Nao[58565:32508349] OCTAudioQueue stop
2016-01-01 13:53:17.648 Nao[58565:32508349] Freed buffers
2016-01-01 13:53:17.649 Nao[58565:32508349] <OCTToxAV: 0x610000046d20>: stop method called
2016-01-01 13:53:17.649 Nao[58565:32508349] <OCTToxAV: 0x610000046d20>: stopped
```

Chuongv:
@stal888 Could you give this a try? Sorry this took me 3 months to fix, I blame it on CS :smile:

stal888:
![1443853841564](https://cloud.githubusercontent.com/assets/774591/14004348/a6f039f4-f116-11e5-9907-5b7893b56fe0.png)

It seems to work.

dvor:
Fixed

PR #162, opened 2016-03-23
Kill OCTSubmanager calls before killing OCTTox

Chuongv:
https://github.com/Antidote-for-Tox/objcTox/issues/154


dvor:
I think it is a good idea to kill _all_ submanagers that we have. Can you add them please?

dvor:
Thanks!

PR #163, opened 2016-03-25
Add test coverage for submanageruser set avatar

Chuongv:
Just a simple test for submanageruser's set avatar method.


Chuongv:
Darn only 0.39% increase 😂

dvor:
Cool

PR #171, opened 2017-09-21
A/V Tweaks for 3g connections

zoff99:


zoff99:
https://github.com/Antidote-for-Tox/toxcore/pull/5

Classes/Private/Manager/Submanagers/OCTSubmanagerCallsImpl.m
```diff
@@ -7,7 +7,7 @@
 #import "OCTTox.h"
 
 const OCTToxAVAudioBitRate kDefaultAudioBitRate = OCTToxAVAudioBitRate48;
-const OCTToxAVVideoBitRate kDefaultVideoBitRate = 2000;
+const OCTToxAVVideoBitRate kDefaultVideoBitRate = 2500;
```
zoff99, line 10: this is OK as default options, since in my opinion most iPhone users will be on 3g network.

but later this should be a **user changeable** setting:

2500 -> for 3g
500 -> for Wifi (or strong LTE network)

dvor: I guess app should change this parameter itself based on connection type.

Classes/Private/Manager/Video/OCTVideoEngine.m
```diff
@@ -46,6 +46,11 @@ - (instancetype)init
 #if ! TARGET_OS_SIMULATOR
     _captureSession = [AVCaptureSession new];
     _captureSession.sessionPreset = AVCaptureSessionPresetMedium;
+
+    if ([_captureSession canSetSessionPreset:AVCaptureSessionPreset640x480])
+    {
+        _captureSession.sessionPreset = AVCaptureSessionPreset640x480;
```
dvor, line 52: Hm, GitHub has confusing review interface, my comment was removed. Anyway, let's try 640x480 for now, I'll play with other options later.

